PR #25, opened 2014-12-04
Update develop scripts due to funds and assets rake task

paoolo:


spec/dummy/bin/prepare-environment
```diff
@@ -9,7 +9,9 @@ pushd $(dirname $(dirname $(dirname ${__dir})))
 popd
 
 pushd $(dirname ${__dir})
+    bundle exec rake assets:precompile
```
paoolo: @mkasztelnik I guees that if there is no ``config.assets.compile = false`` in ``spec/dummy/config/environments/development.rb`` like in ``spec/dummy/config/environments/production.rb`` then after making some commands using ``bundle exec rake`` will automagic recompile assets by default?

mkasztelnik: Yeap

paoolo:
Waiting for +1

.gitignore
```diff
@@ -31,4 +31,7 @@
 /spec/dummy/tmp
 /.project
 nbproject/
-coverage
\ No newline at end of file
+coverage
+
+# Ignore assets
+spec/dummy/public/assets/
```
mkasztelnik: This also can be removed since assets will not be generated using rake task.

mkasztelnik:
After solving issue reported above :+1: from me :smile:

paoolo:
Thanks. Waiting for second +1

nowakowski:
:+1:

PR #68, opened 2015-01-03
Billing flags moved from Appliance to Deployment

nowakowski:
Billing flags (billing_state and prepaid_until) have been moved from Atmosphere::Appliance to Atmosphere::Deployment in order to properly support cases where deployments are dynamically started and shut down during the appliance's lifecycle. The billing service has been refactored to comply with this change.

A set of helper methods has been added to Atmosphere::Appliance in order to mimic the old billing flags on a "best effort" basis - this should prevent breaking backwards compatibility with systems which rely on Atmosphere APIs (CLEW and otherwise).


tbartynski:
I've skimmed through the code. There is a lot of changes but I guess the logic is just moved to other place so it's proved to work correctly.
I'm not sure what is our policy for Hound? Should Hound complains prevent from merging code into master?

Nuanda:
That depends on a specific case and the complaint. In general - use your own judgment ;).

nowakowski:
Not yet ready for review but thanks all the same. :)

mkasztelnik:
Small recommendation - if hound problem is fixed remove hound comment, this allows to clear a little bit PR thread.

nowakowski:
Already cleared the duplicates but okay, will do.

nowakowski:
All done. Deploying on prod would require us to manually set os_family for each ApplianceType but the deploy action should not break anything (all ATs will be set to "Windows").

app/models/atmosphere/appliance.rb
```diff
@@ -129,15 +127,40 @@ def owned_by?(user)
...
+      # Helper method which determines how long this appliance will remain prepaid
+      if deployments.blank?
+        # Invoking this method on an appl whith 0 deployments is a nasal demons
+        # scenario: the question doesn't make sense and neither will the answer.
+        # Returning fixed time is least likely to break the client.
+        result = Time.parse('2000-01-01 12:00')
+      else
+        result = deployments.max_by { |dep| dep.prepaid_until }.prepaid_until
+      end
+      result
```
mkasztelnik: `result` variable doesn't have to be created here. Instead you can use:

```ruby
def prepaid_until
  if deployments.blank?
    Time.parse('2000-01-01 12:00')
  else
    deployments.max_by { |dep| dep.prepaid_until }.prepaid_until
  end
end
```

app/models/atmosphere/appliance.rb
```diff
@@ -129,15 +127,40 @@ def owned_by?(user)
...
+    end
+
+    # This method provided for backward compatibility
+    def billing_state
+      if deployments.any? { |d| d.billing_state == 'prepaid' }
+        'prepaid'
+      else
+        'expired'
```
mkasztelnik: Can we use symbols here?

```ruby
def billing_state
  prepaid? ? :prepaid : :expired
end

def prepaid?
  deployments.any? { |d| d.biling_state.prepaid? }
end
```

app/models/atmosphere/appliance.rb
```diff
@@ -129,15 +127,40 @@ def owned_by?(user)
...
 
     def assign_default_fund
-      # This is a "goalkeeper" method which will assign this appliance to its owner's default fund, if no fund has been specified yet.
-      # It is provided to ensure compatibility with old APIs of Atmosphere which do not request funds to be specified when instantiating appliances
+      # This is a "goalkeeper" method which will assign this appliance
+      # to its owner's default fund, if no fund has been specified yet.
+      # It is provided to ensure compatibility with old APIs of Atmosphere
+      # which do not request funds to be specified when instantiating appliances
       # Once the APIs have been updated, this method will be deprecated and should be removed.
       if self.fund.blank?
```
mkasztelnik: `self` is redundant for getter methods, it is required only for setters, so it can be changed into:

```ruby
if fund.blank?
  self.fund = appliance_set.user.default_fund
  ...
```

app/models/atmosphere/appliance.rb
Nuanda: Regarding symbols, we need to be careful. First - afaik ActiveRecord does not automatically translate string field values into ruby symbols, so 'prepaid' != :prepaid. We, however, use enumerize to kind of bypass that. So we can't use symbols here unless we change carefully the enumerize use in line 77.

app/models/atmosphere/virtual_machine_flavor.rb
```diff
@@ -60,6 +64,48 @@ class VirtualMachineFlavor < ActiveRecord::Base
...
+    def set_hourly_cost_for(os_family, cost)
+      incarnation = virtual_machine_flavor_os_families.select{|vmf_osf| vmf_osf.os_family == os_family}.first
+      if incarnation.is_a?(Atmosphere::VirtualMachineFlavorOSFamily)
+        incarnation.hourly_cost = cost
+        incarnation.save
+      else
+        incarnation = Atmosphere::VirtualMachineFlavorOSFamily.new(virtual_machine_flavor: self, os_family: os_family,
+          hourly_cost: cost)
+        incarnation.save
+      end
```
tbartynski: Move incarnation.save out if since it is present in both if branches.

app/services/atmosphere/billing_service.rb
```diff
@@ -72,48 +84,45 @@ def self.bill_appliance(appl, billing_time, message, apply_prepayment = true)
...
+        # If this VM has at least one deployment whose funding policy states 'no action' then do nothing.
+        elsif vm.deployments.select {|dep| dep.appliance.fund.termination_policy == 'no_action'}.count > 0
           # Do nothing
-          Rails.logger.debug("Leaving VM #{vm.id} unchanged -- there are no prepaid appliances using it but the corresponding fund termination policy is 'no_action'.")
-        # If this VM has at least one appliance whose funding policy states 'suspend' then shut down the VM without deleting it
-        elsif vm.appliances.select {|appl| appl.fund.termination_policy == 'suspend'}.count > 0
+          Rails.logger.debug("Leaving VM #{vm.id} unchanged -- there are no prepaid deployments using it but the corresponding fund termination policy is 'no_action'.")
+        # If this VM has at least one deployment whose funding policy states 'suspend' then shut down the VM without deleting it
+        elsif vm.deployments.select {|dep| dep.appliance.fund.termination_policy == 'suspend'}.count > 0
           # TODO: Ask TB how to suspend this VM
```
tbartynski: See VirtualMachine suspend method.

app/services/atmosphere/billing_service.rb
```diff
@@ -131,12 +140,25 @@ def self.can_afford_vm?(appliance, vm)
...
+        billable_time = self.calculate_billable_time(deployments.first, Time.now.utc, true)
+      else
+        # If this VM is not assigned to appliance then just assume def prepayment interval
+        billable_time = @appliance_prepayment_interval/3600 # Time in hours
+      end
+      hourly_cost = vm.virtual_machine_flavor.get_hourly_cost_for(appliance.appliance_type.os_family)
+      if hourly_cost.blank? # Will happen when there is a mismatch between AT ostype and flavor ostype
+        return false
+      end
+      amt_due = ((billable_time*hourly_cost)/(vm.appliances.count+1)).round
```
tbartynski: Why is there vm.appliances.count + 1?

nowakowski: The +1 represents the (still hypothetical) new appliance for which the charge is returned. Basically, the VM is shared by n appliances and we're asking how much it would cost to assign another appliance to it.

spec/factories/appliance_types.rb
```diff
@@ -23,6 +23,13 @@
       virtual_machine_templates { [ build(:virtual_machine_template)] }
     end
 
+    # If no os_family has been specified by the requestor, bind this appliance to the first available os_family
+    after(:build) do |atype|
+      if atype.os_family.blank?
+        atype.os_family = Atmosphere::OSFamily.first
```
tbartynski: Are we sure that Atmosphere::OSFamily.first is not nil?

nowakowski: Yes, the migration sets up 1 record automatically. With a blank OSFamily table nothing in the system will work anyway.

spec/services/atmosphere/billing_service_spec.rb
```diff
@@ -139,7 +165,7 @@
       expect(Atmosphere::BillingService.can_afford_flavor?(standalone_zus_appl, cheap_flavor)).to eq false
       expect(Atmosphere::BillingService.can_afford_flavor?(amber_gold_appl, expensive_flavor)).to eq true
 
-      # aplien_appl cannot afford anything because its fund is not bound to ComputeSite cs
+      # alien_appl cannot afford anything because its fund is not bound to ComputeSite cs
       expect(Atmosphere::BillingService.can_afford_flavor?(alien_appl, cheap_flavor)).to eq false
       expect(Atmosphere::BillingService.can_afford_flavor?(alien_appl, free_flavor)).to eq false
     end
```
tbartynski: This test should be split into 3 examples that can be within one context. Shared variables should be defined in the context scope.

spec/services/atmosphere/billing_service_spec.rb
```diff
@@ -148,6 +174,11 @@
       expensive_vm = create(:virtual_machine, compute_site: cs, source_template: tmpl_of_not_shareable_at, virtual_machine_flavor: expensive_flavor)
       cheap_vm = create(:virtual_machine, compute_site: cs, source_template: tmpl_of_not_shareable_at, virtual_machine_flavor: cheap_flavor)
       free_vm = create(:virtual_machine, compute_site: cs, source_template: tmpl_of_not_shareable_at, virtual_machine_flavor: free_flavor)
+      shareable_vm = create(:virtual_machine, compute_site: cs, source_template: tmpl_of_shareable_at, virtual_machine_flavor: free_flavor)
+
+      free_flavor.virtual_machine_flavor_os_families.first.update_attribute(:hourly_cost, 0)
+      cheap_flavor.virtual_machine_flavor_os_families.first.update_attribute(:hourly_cost, 10)
+      expensive_flavor.virtual_machine_flavor_os_families.first.update_attribute(:hourly_cost, 100)
```
tbartynski: Flavor definitions can be moved to above context and be reused. Consider testing only one aspect in one example.

spec/services/atmosphere/billing_service_spec.rb
```diff
@@ -162,6 +193,41 @@
...
+    let!(:linux_appliance) { create(:appliance, fund: poor_fund, appliance_set: wf,
+      appliance_type: linux_appl_type, appliance_configuration_instance: config_inst,
+      virtual_machines: [not_shareable_vm2]) }
+
+    it 'can or cannot afford flavors depending on os_family' do
+
+      windows_flavor.os_families = [windows_osfamily]
+      linux_flavor.os_families = [linux_osfamily]
+      windows_flavor.save
+      linux_flavor.save
```
tbartynski: Can this be moved to let declarations above?

spec/services/atmosphere/billing_service_spec.rb
```diff
@@ -162,6 +193,41 @@
...
+
+    let!(:windows_appliance) { create(:appliance, fund: poor_fund, appliance_set: wf,
+      appliance_type: windows_appl_type, appliance_configuration_instance: config_inst,
+      virtual_machines: [not_shareable_vm1]) }
+
+    let!(:linux_appliance) { create(:appliance, fund: poor_fund, appliance_set: wf,
+      appliance_type: linux_appl_type, appliance_configuration_instance: config_inst,
+      virtual_machines: [not_shareable_vm2]) }
+
+    it 'can or cannot afford flavors depending on os_family' do
```
tbartynski: Again, consider testing 1 thing in 1 example.

app/models/atmosphere/appliance.rb
mkasztelnik: @Nuanda are you referring [this line](https://github.com/dice-cyfronet/atmosphere/blob/billing_in_deployments/app/models/atmosphere/deployment.rb#L22)? If yes then my `prepaid?` method is still valid. According to `prepaid` and `expired` in `billing_state` you are absolutely right - I was not checking how it is used - my fault.

app/models/atmosphere/deployment.rb
```diff
@@ -10,10 +10,24 @@
...
       class_name: 'Atmosphere::Appliance'
 
     belongs_to :virtual_machine,
       class_name: 'Atmosphere::VirtualMachine'
+
+    validates :billing_state, presence: true
+
+    enumerize :billing_state, in: ["initial", "prepaid", "expired", "error"], predicates: true
+
+    after_create :initial_billing
```
mkasztelnik: `initial_billing` method is changing fields values and this callback is invoked. According to the [documentation](http://api.rubyonrails.org/classes/ActiveRecord/Callbacks.html) `after_create` callback is invoked after object is saved into DB. Maybe `before_create` callback should be used here?

app/models/atmosphere/virtual_machine_flavor.rb
```diff
@@ -60,6 +64,48 @@ class VirtualMachineFlavor < ActiveRecord::Base
 
     scope :active, -> { where(active: true) }
 
+    # Assumes only 1 vmf_osf for a specific os_family
+    def get_hourly_cost_for(os_family)
+      incarnation = virtual_machine_flavor_os_families.select{|vmf_osf| vmf_osf.os_family == os_family}.first
```
mkasztelnik: You can use DB query here:

```ruby
virtual_machine_flavor_os_families.find_by(os_family: os_family)
```

app/models/atmosphere/virtual_machine_flavor.rb
```diff
@@ -60,6 +64,48 @@ class VirtualMachineFlavor < ActiveRecord::Base
...
+      if incarnation.is_a?(Atmosphere::VirtualMachineFlavorOSFamily)
+        incarnation.hourly_cost
+      else
+        nil #GIGO
+      end
```
mkasztelnik: This if/else can be simplified into:

```ruby
incarnation && incarnation.hourly_cost
```

app/models/atmosphere/virtual_machine_flavor.rb
mkasztelnik: It can be simplified as follow:

```ruby
def set_hourly_cost_for(os_family, cost)
  incarnation = virtual_machine_flavor_os_families.find_or_initialize_by(os_family: os_family)
  incarnation.hourly_cost = cost
  incarnation.save
end
```

app/models/atmosphere/virtual_machine_flavor.rb
```diff
@@ -60,6 +64,48 @@ class VirtualMachineFlavor < ActiveRecord::Base
...
+    end
+
+    # Provides backward compatibility with old versions of the GUI
+    def hourly_cost
+      if virtual_machine_flavor_os_families.blank?
+        nil
+      else
+        virtual_machine_flavor_os_families.max_by(&:hourly_cost).hourly_cost
+      end
+    end
```
mkasztelnik: It can be simplified as follow:

```ruby
def hourly_cost
  virtual_machine_flavor_os_families &&
    virtual_machine_flavor_os_families.maximum(:hourly_cost)
end
```

app/models/atmosphere/virtual_machine_flavor.rb
```diff
@@ -60,6 +64,48 @@ class VirtualMachineFlavor < ActiveRecord::Base
...
+
+    # Returns a full cost map for this flavor (depending on os_family)
+    def cost_map
+      result = {}
+
+      virtual_machine_flavor_os_families.each do |f|
+        result[f.os_family.os_family_name] = f.hourly_cost
+      end
+      result
+    end
```
mkasztelnik: It can be simplified:

```ruby
def cost_map
   Hash[virtual_machine_flavor_os_families.map {|f| [f.os_family.os_family_name, f.hourly_cost] }]
end
```

or

```ruby
def cost_map
  flavor.virtual_machine_flavor_os_families.each_with_object({}) do |f, hash|
    hash[f.os_family.os_family_name] = f.hourly_cost
  end
end
```

app/models/atmosphere/virtual_machine_flavor_os_family.rb
```diff
@@ -0,0 +1,15 @@
...
+module Atmosphere
+  class VirtualMachineFlavorOSFamily < ActiveRecord::Base
+    belongs_to :virtual_machine_flavor,
+      class_name: 'Atmosphere::VirtualMachineFlavor'
+
+    belongs_to :os_family,
+      class_name: 'Atmosphere::OSFamily'
+
+    #validates :hourly_cost,
+    #  numericality: { greater_than_or_equal_to: 0 }
```
mkasztelnik: If this validation is not necessary please remote lines 11 and 12

app/providers/atmosphere/optimization_strategy/default.rb
```diff
@@ -80,7 +80,7 @@ def select
                   f.cpu >= min_cpu &&
                   f.hdd >= min_hdd
                 end
-              ) {|f| f.hourly_cost}
+              ) {|f| f.get_hourly_cost_for(tmpl.appliance_type.os_family)}
```
mkasztelnik: Maybe `get_hourly_cost_for` can take `tmpl` as an argument and do `tmpl.appliance_type.os_family` internally?

app/serializers/atmosphere/virtual_machine_flavor_serializer.rb
```diff
@@ -5,7 +5,23 @@ class VirtualMachineFlavorSerializer < ActiveModel::Serializer
...
+    end
+
+    # Returns a full cost map for this flavor (depending on os_family)
+    def cost_map
+      result = {}
+      object.virtual_machine_flavor_os_families.each do |f|
+        result[f.os_family.os_family_name] = f.hourly_cost
+      end
+      result
+    end
```
mkasztelnik: `cost_map` is already defined in `VirtualMachineFlavor`, you can use it instead of duplicating the code.

app/services/atmosphere/billing_service.rb
```diff
@@ -20,45 +20,57 @@ def self.bill_all_appliances
...
+    end
+
+
+    def self.bill_deployment(dep, billing_time, message, apply_prepayment = true)
+      appl = dep.appliance # Shorthand
+      if dep.prepaid_until.blank?
+        # This should not happen - it means that we do not know when this deployment will run out of funds.
+        # Therefore we cannot meaningfully bill it again and must return an error.
+        Rails.logger.error("Unable to determine current payment validity date for deployment #{dep.id} belonging to appliance #{appl.id}. Skipping.")
+        BillingLog.new(timestamp: Time.now.utc, user: appl.appliance_set.user, appliance: appl.id.to_s+'---'+(appl.name.blank? ? 'unnamed_appliance' : appl.name), fund: appl.fund.name, message: "Unable to determine current payment validity date for deployment.", actor: "bill_appliance", amount_billed: 0).save
```
mkasztelnik: You can use `"#{val1}--#{val2}" construction here

app/services/atmosphere/billing_service.rb
```diff
@@ -20,45 +20,57 @@ def self.bill_all_appliances
...
-        appl.billing_state = "error"
-        appl.save
-        raise Atmosphere::BillingException.new(message: "Unable to determine current payment validity date for appliance #{appl.id}")
+      appl.deployments.each do |dep|
+        bill_deployment(dep, billing_time, message, apply_prepayment)
+      end
+    end
+
+
+    def self.bill_deployment(dep, billing_time, message, apply_prepayment = true)
```
mkasztelnik: Would it be possible to simplify this method by splitting it into set of smaller methods? Maybe it can be moved into separate class?

mkasztelnik:
`billing_service.rb` code is illegible using github file browser - [take a look here](https://github.com/dice-cyfronet/atmosphere/blob/billing_in_deployments/app/services/atmosphere/billing_service.rb). Can you reformat it and keep more/less 80 chars line long?

app/services/atmosphere/billing_service.rb
```diff
@@ -20,45 +20,57 @@ def self.bill_all_appliances
...
+        billing_interval = (billing_time - dep.prepaid_until) # This will return time in seconds
         if billing_interval < 0
           # The appliance is still prepaid - nothing to be done.
-          Rails.logger.debug("Appliance #{appl.id} is prepaid until #{appl.prepaid_until}. Nothing to be done.")
+          Rails.logger.debug("Deployment #{dep.id} belonging to appliance #{appl.id} is prepaid until #{dep.prepaid_until}. Nothing to be done.")
         else
-          # Bill the hell out of this appliance! :)
-          Rails.logger.debug("Billing appliance #{appl.id} for #{(billing_interval/3600)} hours of use.")
+          # Bill the hell out of this deployment! :)
+          Rails.logger.debug("Billing deployment #{dep.id} belonging to appliance #{appl.id} for #{(billing_interval/3600)} hours of use.")
```
mkasztelnik: `()` is not required in `#{(billing_interval/3600)}`

app/services/atmosphere/billing_service.rb
```diff
@@ -20,45 +20,57 @@ def self.bill_all_appliances
...
-              appl.billing_state = "prepaid"
+              dep.prepaid_until = billing_time+(apply_prepayment ? @appliance_prepayment_interval : 0)
+              dep.billing_state = "prepaid"
               appl.save
-              if appl.errors.blank?
+              dep.save
+              if dep.errors.blank?
                 # Write success to log.
                 BillingLog.new(timestamp: Time.now.utc, user: appl.appliance_set.user, appliance: appl.id.to_s+'---'+(appl.name.blank? ? 'unnamed_appliance' : appl.name), fund: appl.fund.name, message: message, actor: "bill_appliance", amount_billed: amount_due).save
               else
```
mkasztelnik: In else statement you are raising exception. Will it rollback whole transaction? If yes then error billing log will not be saved in db, if no than we can have data inconsistency.

app/services/atmosphere/billing_service.rb
```diff
@@ -72,48 +84,45 @@ def self.bill_appliance(appl, billing_time, message, apply_prepayment = true)
...
-      billable_time = self.calculate_billable_time(appl, billing_time, apply_prepayment)
+    # - This is usually true, but will be false when calculating final charge for a deployment which is being shut down.
+    def self.calculate_charge_for_deployment(dep, billing_time, apply_prepayment = true)
+      # By default, the billing process should extend the validity of the deployment until now+1.hour.
+      billable_time = self.calculate_billable_time(dep, billing_time, apply_prepayment)
 
-      amount_due = 0
+      vm = dep.virtual_machine # Shorthand
+      # Find out how many appliances are using this VM and split costs equally
+      os_family = dep.appliance.appliance_type.os_family
```
mkasztelnik: Please try to apply [Law of Diemeter](http://en.wikipedia.org/wiki/Law_of_Demeter) when possible. Here inside `appliance` `os_family` method can be added.

app/services/atmosphere/billing_service.rb
```diff
@@ -72,48 +84,45 @@ def self.bill_appliance(appl, billing_time, message, apply_prepayment = true)
...
-        # Find out how many appliances are using this VM and split costs equally
-        hourly_charge = (vm.virtual_machine_flavor.hourly_cost/vm.appliances.count).round #TODO: figure out if we should limit this to appliances with billing_state == :prepaid
-        Rails.logger.debug("Calculated hourly charge for using VM #{vm.id} is #{hourly_charge}. This VM currently has #{vm.appliances.count} appliances using it.")
-        charge = (hourly_charge*billable_time).round
-        amount_due += charge
-      end
-      # Return amount due
-      amount_due
+      # Return anticipated charge
+      charge
```
mkasztelnik: This return statement is redundant. You can remove `charge` variable and simply return

```
(hourly_charge*billable_time).round
```

app/services/atmosphere/billing_service.rb
```diff
@@ -72,48 +84,45 @@ def self.bill_appliance(appl, billing_time, message, apply_prepayment = true)
...
     end
 
-    # This scans for :expired appliances and figures out what to do with the underlying VMs, following fund policies
+    # This scans for :expired deployments and figures out what to do with the underlying VMs, following fund policies
     def self.apply_funding_policy
       VirtualMachine.where(managed_by_atmosphere: true).all.each do |vm|
```
mkasztelnik:
```ruby
VirtualMachine.manageable.each do |vm|
  ...
```

app/services/atmosphere/billing_service.rb
```diff
@@ -72,48 +84,45 @@ def self.bill_appliance(appl, billing_time, message, apply_prepayment = true)
...
-        # If this VM has at least one prepaid appliance, it must not be touched. For safety's sake, we will also not touch appliances whose billing state is flagged as erroneous
-        if vm.appliances.select {|appl| ['prepaid', 'error'].include? appl.billing_state}.count > 0
+        # If this VM has at least one prepaid deployment, it must not be touched. For safety's sake, we will also not touch deployments whose billing state is flagged as erroneous
+        if vm.deployments.select {|dep| ['prepaid', 'error'].include? dep.billing_state}.count > 0
```
mkasztelnik: All this complex conditions (in if/elseifs) can be externalized into dedicated methods with meaningful names.

app/services/atmosphere/billing_service.rb
```diff
@@ -131,12 +140,25 @@ def self.can_afford_vm?(appliance, vm)
       if appliance.fund.blank?
         raise Atmosphere::BillingException.new(message: "can_afford_vm? invoked on an appliance (with id #{appliance.id}) which has no fund assigned. Unable to proceed.")
       end
-      billable_time = self.calculate_billable_time(appliance, Time.now.utc, true)
-      amt_due = (billable_time*(vm.virtual_machine_flavor.hourly_cost)/(vm.appliances.count+1)).round
+
+      # Check whether the target VM is already assigned to the appliance via deployments
+      # If so, respect the prepayment period which *may* be present for this deployment
+      deployments = appliance.deployments.select {|dep| dep.virtual_machine == vm}
```
mkasztelnik:
```ruby
deployments = appliance.deployments.where(virtual_machine_id: vm.id)
```

BTW I believe that it is not possible to have more than 1 deployment relation between appliance and vm, then we can use:

```ruby
deployments = appliance.deployments.find_by(virtual_machine_id: vm.id)
```

app/services/atmosphere/billing_service.rb
```diff
@@ -131,12 +140,25 @@ def self.can_afford_vm?(appliance, vm)
...
+        return false
+      end
+      amt_due = ((billable_time*hourly_cost)/(vm.appliances.count+1)).round
       if amt_due <= (appliance.fund.balance-appliance.fund.overdraft_limit) and appliance.fund.compute_sites.include? vm.compute_site
-        true
+        return true
       else
-        false
+        return false
       end
```
mkasztelnik: This if else is redundant. You can simply use:

```
amt_due <= (appliance.fund.balance-appliance.fund.overdraft_limit) && 
  appliance.fund.compute_sites.include?(vm.compute_site)
```

What is more be careful when using `and` instead of  `&&` it has [lower precedence](http://stackoverflow.com/questions/1426826/difference-between-and-and-in-ruby) than `&&`

mkasztelnik: Order in this method can be changed. At the beginning you can check for "mismatch between AT ostype and flavor ostype", only when there everything is ok you can calculate `billable_time` and final condition.

app/services/atmosphere/billing_service.rb
```diff
@@ -146,12 +168,16 @@ def self.can_afford_flavor?(appliance, flavor)
...
-      billable_time = self.calculate_billable_time(appliance, Time.now.utc, true)
-      amt_due = (billable_time*(flavor.hourly_cost)).round
+      billable_time = @appliance_prepayment_interval/3600 # Time in hours
+      hourly_cost = flavor.get_hourly_cost_for(appliance.appliance_type.os_family)
+      if hourly_cost.blank? # Will happen when there is a mismatch between AT ostype and flavor ostype
+        return false
+      end
+      amt_due = (billable_time*(hourly_cost)).round
       if amt_due <= (appliance.fund.balance-appliance.fund.overdraft_limit) and appliance.fund.compute_sites.include? flavor.compute_site
```
mkasztelnik: Similar code as in previous method - can be extracted into separate method.

db/migrate/20150103133015_move_prepaid_until_to_deployments.rb
```diff
@@ -0,0 +1,45 @@
...
+class MovePrepaidUntilToDeployments < ActiveRecord::Migration
+  def up
+    add_column :atmosphere_deployments, :billing_state, :string, null:false, default:"prepaid"
+
+    # Time for some good old-fashioned SQL because the Rails ORM model is too constraining to support current timestamp as default datetime
+    execute "ALTER TABLE atmosphere_deployments ADD COLUMN prepaid_until TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP"
+
+    # Now we need to iterate over existing Appliances and rewrite the relevant data in order to maintain appliance info
+    Atmosphere::Appliance.find_each do |a|
+      a.deployments.each do |dep|
```
mkasztelnik: n+1 queries here. You can [eager load](http://guides.rubyonrails.org/active_record_querying.html#eager-loading-associations) deployments to avoid additional db queries.

db/migrate/20150103133015_move_prepaid_until_to_deployments.rb
```diff
@@ -0,0 +1,45 @@
...
+    remove_column :atmosphere_appliances, :prepaid_until
+
+  end
+
+  def down
+    add_column :atmosphere_appliances, :billing_state, :string, null:false, default:"prepaid"
+    execute "ALTER TABLE atmosphere_appliances ADD COLUMN prepaid_until TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP"
+
+    Atmosphere::Appliance.find_each do |a|
+      dep = a.deployments.first
```
mkasztelnik: n+1 queries here. You can eager load deployments to avoid additional db queries.

db/migrate/20150113100404_add_osfamilies.rb
```diff
@@ -0,0 +1,63 @@
...
+      t.belongs_to :os_family
+    end
+
+    add_reference :atmosphere_appliance_types, :atmosphere_os_families, index: true
+
+    # Spawn two records and rewrite existing ATypes to bind to a specific os_family
+    os_windows = Atmosphere::OSFamily.new(os_family_name: 'Windows')
+    os_linux = Atmosphere::OSFamily.new(os_family_name: 'Linux')
+    os_windows.save
+    os_linux.save
```
mkasztelnik: You can simplify it:

```ruby
os_windows = Atmosphere::OSFamily.create(os_family_name: 'Windows')
Atmosphere::OSFamily.create(os_family_name: 'Linux')
```

db/migrate/20150113100404_add_osfamilies.rb
```diff
@@ -0,0 +1,63 @@
...
+    # Rewrite all existing ATs to use the 'windows' OS (correct later as necessary)
+
+    Atmosphere::ApplianceType.find_each do |atype|
+      atype.os_family = os_windows
+      atype.save
+    end
+
+    # Assign an os_family to each existing flavor and rewrite cost
+    Atmosphere::VirtualMachineFlavor.find_each do |flavor|
+      vmf_osf = Atmosphere::VirtualMachineFlavorOSFamily.new(virtual_machine_flavor: flavor,
```
mkasztelnik: Similar as above: `create` instead of `new` and then `save`

db/migrate/20150113100404_add_osfamilies.rb
```diff
@@ -0,0 +1,63 @@
...
+        os_family: Atmosphere::OSFamily.first,
+        hourly_cost: flavor.hourly_cost)
+      vmf_osf.save
+    end
+
+    # Note: a separate migration will be needed to remove Atmosphere::VirtualMachineFlavor.hourly_cost
```
mkasztelnik: Hmm I don't understand this note. In line 43 you are removing this field.

spec/factories/appliance_types.rb
mkasztelnik: If this is required always than `OSFamily` creation should be also added into `seed.rb` in every project which uses atmosphere. `rails generate atmosphere:install` task is a good place to add code responsible for adding additional seeds.

What is more if `OSFamily` is required than maybe we should have validation to not allow to remove all of them from db?

spec/providers/atmosphere/optimization_strategy/manual_spec.rb
```diff
@@ -11,13 +11,20 @@
   let(:cs) { create(:compute_site, active: true, funds: [fund]) }
   let(:at) { create(:filled_appliance_type, visible_to: 'all', preference_disk: 0) }
   let!(:tmpl) { create(:virtual_machine_template, appliance_type: at, compute_site: cs) }
-  let!(:fl1) { create(:virtual_machine_flavor, cpu: 1, memory: 512, compute_site: cs, active: true, hourly_cost: 1) }
-  let!(:fl2) { create(:virtual_machine_flavor, cpu: 2, memory: 1024, compute_site: cs, active: true, hourly_cost: 2) }
+  let!(:fl1) { create(:virtual_machine_flavor, cpu: 1, memory: 512, compute_site: cs, active: true) }
+  let!(:fl2) { create(:virtual_machine_flavor, cpu: 2, memory: 1024, compute_site: cs, active: true) }
```
mkasztelnik: Please take a look that `fl1` and `fl2` are not used in second test, but these flavors will be created in db. I know that here it is only 2 additional inserts but "ziarnko do ziarnka a zbierze się miarka" :smile: 

Try to refactor it in such a way that every spec will have only required db object created for every spec.

spec/providers/atmosphere/optimization_strategy/manual_spec.rb
```diff
@@ -11,13 +11,20 @@
...
   let(:cfg_tmpl) { create(:appliance_configuration_template, appliance_type: at)}
   let(:user) { create(:user) }
   let(:as) { create(:appliance_set, user: user)}
 
   it 'return templates and flavors for each defined vm in opt policy params' do
+
+    fl1.set_hourly_cost_for(at.os_family, 5)
+    fl2.set_hourly_cost_for(at.os_family, 10)
```
mkasztelnik: Why setting cost here? Manual optimization takes a look at the beginning at `cpu` and `ram`, cost at the end - so here different costs should not do any difference.

spec/requests/atmosphere/api/clew_spec.rb
```diff
@@ -42,6 +42,9 @@
       it 'returns appropriate response' do
         get api("/clew/appliance_instances", user)
         expect(response.status).to eq 200
+
+        appl.reload
```
mkasztelnik: Why `appl` need to be reloaded here?

spec/requests/atmosphere/api/virtual_machine_flavors_spec.rb
```diff
@@ -33,6 +33,7 @@
       it 'returns 200' do
         get api('/virtual_machine_flavors', user)
         expect(response.status).to eq 200
+
```
mkasztelnik: Extra empty line

spec/serializers/atmosphere/virtual_machine_flavor_serializer_spec.rb
```diff
@@ -1,38 +1,66 @@
 require 'rails_helper'
 
 describe Atmosphere::VirtualMachineFlavorSerializer do
-  it 'sets active flag to true for flavor on active compute site' do
-    flavor = flavor_on_cs(cs_active: true, active: true)
 
-    result = serialize(flavor)
+  context 'basic tests' do
```
mkasztelnik: Very strange name for the context. What added value it brings?

spec/serializers/atmosphere/virtual_machine_flavor_serializer_spec.rb
```diff
@@ -1,38 +1,66 @@
...
-  end
+      result = serialize(flavor)
 
-  it 'sets active flag to true for flavor on inactive compute site' do
-    flavor = flavor_on_cs(cs_active: false, active: true)
+      expect(result['virtual_machine_flavor']['active']).to be_truthy
+    end
 
-    result = serialize(flavor)
+    it 'sets active flag to true for flavor on inactive compute site' do
```
mkasztelnik: sets active flag to _true_ for flavor on inactive compute site => sets active flag to _false_ for flavor on inactive compute site

mkasztelnik: Maybe better name for this and previous spec: "is active/inactive on active/inactive compute site"

spec/serializers/atmosphere/virtual_machine_flavor_serializer_spec.rb
```diff
@@ -1,38 +1,66 @@
...
+      flavor = flavor_on_cs(cs_active: false, active: true)
 
-    expect(result['virtual_machine_flavor']['active']).to be_falsy
-  end
+      result = serialize(flavor)
+
+      expect(result['virtual_machine_flavor']['active']).to be_falsy
+    end
+
+    it 'sets active falg to false for flavor without compute site' do
```
mkasztelnik: Most probably these are my mistakes :sob:: falg -> flag

spec/services/atmosphere/billing_service_spec.rb
```diff
@@ -40,7 +45,9 @@
 
     it 'throws exception for malformed appliance' do
       appl = create(:appliance, fund: nil, appliance_set: wf, appliance_type: not_shareable_appl_type, appliance_configuration_instance: config_inst, virtual_machines: [not_shareable_vm1])
-      appl.update_column(:fund_id, nil)
+
+      appl.update_columns(fund_id: nil)
```
mkasztelnik: Can you explain why you are nilling this value with callbacks skipped? If there is dedicated callback/validation which guarantee that appliance without fund cannot exist maybe this spec can be removed?

spec/services/atmosphere/billing_service_spec.rb
```diff
@@ -74,7 +91,8 @@
 
       Atmosphere::BillingService.bill_appliance(appl, Time.now.utc, "mockup billing", true)
 
-      expect(appl.billing_state).to eq("prepaid")
+      appl.reload
+      expect(appl.deployments.first.billing_state).to eq("prepaid")
```
mkasztelnik: I see 3 test cases here. Can you split it into 3 separate specs?

spec/services/atmosphere/billing_service_spec.rb
```diff
@@ -83,22 +101,26 @@
...
       Atmosphere::BillingService.bill_appliance(appl2, Time.now.utc, "mock billing", true)
       appl2.reload
-      expect(appl1.amount_billed).to eq appl1.virtual_machines.first.virtual_machine_flavor.hourly_cost
-      expect(appl2.amount_billed).to eq (appl2.virtual_machines.first.virtual_machine_flavor.hourly_cost*0.5).round
+      expect(appl1.amount_billed).to eq appl1.virtual_machines.first.virtual_machine_flavor.virtual_machine_flavor_os_families.first.hourly_cost
+      expect(appl2.amount_billed).to eq (appl2.virtual_machines.first.virtual_machine_flavor.virtual_machine_flavor_os_families.first.hourly_cost*0.5).round
 
       cs_fund.reload
-      expect(cs_fund.balance).to eq original_balance - (appl2.virtual_machines.first.virtual_machine_flavor.hourly_cost*1.5).round
+      expect(cs_fund.balance).to eq original_balance - (appl2.virtual_machines.first.virtual_machine_flavor.virtual_machine_flavor_os_families.first.hourly_cost*1.5).round
```
mkasztelnik: It seems to me like here logic from billing is repeated - so bug here and in billing logic can be repeated. Maybe you can calculate on paper what balance it should be with given precondition and use calculated number?

spec/services/atmosphere/billing_service_spec.rb
```diff
@@ -126,6 +148,10 @@
     let!(:alien_appl) { create(:appliance, fund: alien_fund, appliance_set: wf, appliance_type: not_shareable_appl_type, appliance_configuration_instance: config_inst, virtual_machines:[])}
 
     it 'can or cannot afford flavors' do
```
mkasztelnik: It seems like this spec and the one below tests to much. Please try to test only one thing in one spec - it will allow other developers to understand behaviour easily.

spec/services/atmosphere/billing_service_spec.rb
```diff
@@ -139,7 +165,7 @@
       expect(Atmosphere::BillingService.can_afford_flavor?(standalone_zus_appl, cheap_flavor)).to eq false
       expect(Atmosphere::BillingService.can_afford_flavor?(amber_gold_appl, expensive_flavor)).to eq true
 
-      # aplien_appl cannot afford anything because its fund is not bound to ComputeSite cs
+      # alien_appl cannot afford anything because its fund is not bound to ComputeSite cs
```
mkasztelnik: The note above is a perfect description of new spec :)

spec/services/atmosphere/billing_service_spec.rb
```diff
@@ -162,6 +193,41 @@
...
 
+  context 'os_families' do
+    let(:windows_flavor) { create(:virtual_machine_flavor, compute_site: cs)}
+    let(:linux_flavor) { create(:virtual_machine_flavor, compute_site: cs)}
+
+    let(:poor_fund) { create(:fund, balance: 75, overdraft_limit: 0, compute_sites: [cs] )}
+
+    let(:config_inst) { create(:appliance_configuration_instance) }
+
+    let!(:windows_appl_type) { create(:not_shareable_appliance_type, os_family: windows_osfamily) }
```
mkasztelnik: `let`s are reasonable when reusing in many specs, here they are used only for one spec,  instead of defining `let` you can use local variables.

spec/services/atmosphere/billing_service_spec.rb
```diff
@@ -236,11 +311,11 @@
...
-      expect(appl1.billing_state).to eq "prepaid"
-      expect(appl2.billing_state).to eq "prepaid"
-      expect(appl3a.billing_state).to eq "prepaid"
-      expect(appl3b.billing_state).to eq "prepaid"
-      expect(expiring_appl.billing_state).to eq "expired"
+      expect(appl1.deployments.first.billing_state).to eq "prepaid"
+      expect(appl2.deployments.first.billing_state).to eq "prepaid"
+      expect(appl3a.deployments.first.billing_state).to eq "prepaid"
+      expect(appl3b.deployments.first.billing_state).to eq "prepaid"
+      expect(expiring_appl.deployments.first.billing_state).to eq "expired"
```
mkasztelnik: I see 5 specs here - please split it. As @tbartynski mentioned one spec should test only one thing.

spec/services/atmosphere/billing_service_spec.rb
```diff
@@ -255,15 +330,20 @@
       # Standard non-shared appliance
       appl1 = create(:appliance, fund: cs_fund, appliance_set: wf, appliance_type: not_shareable_appl_type, appliance_configuration_instance: config_inst1, virtual_machines: [not_shareable_vm1])
 
-      cost_unit = appl1.virtual_machines.first.virtual_machine_flavor.hourly_cost # Cost per 1h of use for a single (full) VM
+      vmf = appl1.virtual_machines.first.virtual_machine_flavor
+      vmf.set_hourly_cost_for(Atmosphere::OSFamily.first, 10)
+      vmf.reload
```
mkasztelnik: Why reload is needed every time you are setting hourly cost?

spec/services/atmosphere/billing_service_spec.rb
```diff
@@ -255,15 +330,20 @@
...
+
+      cost_unit = appl1.virtual_machines.first.virtual_machine_flavor.virtual_machine_flavor_os_families.first.hourly_cost # Cost per 1h of use for a single (full) VM
 
       Atmosphere::BillingService.bill_appliance(appl1, Time.now.utc, "mock billing", true)
 
       cs_fund.reload
       expect(cs_fund.balance).to eq original_balance - cost_unit
```
mkasztelnik: Are you testing final bling or normal billing here?

spec/services/atmosphere/billing_service_spec.rb
```diff
@@ -255,15 +330,20 @@
...
 
+
       # Advance the clock by 1.5 hours and destroy appl1
-      appl1.prepaid_until = Time.now - 30.minutes
+      appl1.deployments.first.prepaid_until = Time.now - 30.minutes
       appl1.save
 
       begin
```
mkasztelnik: This rescue is suspicious - conflict exception should not be thrown by fog while destroying vm

spec/services/atmosphere/optimizer_spec.rb
```diff
@@ -409,19 +409,27 @@
...
           }
         end
 
         it 'selects flavor with more ram if prices are equal' do
           biggest_os_flavor = openstack.virtual_machine_flavors.max_by {|f| f.memory}
-          optimal_flavor = create(:virtual_machine_flavor, memory: biggest_os_flavor.memory + 256, cpu: biggest_os_flavor.cpu, hdd: biggest_os_flavor.hdd, hourly_cost: biggest_os_flavor.hourly_cost, compute_site: amazon)
+          optimal_flavor = create(:virtual_machine_flavor, memory: biggest_os_flavor.memory + 256, cpu: biggest_os_flavor.cpu, hdd: biggest_os_flavor.hdd, compute_site: amazon)
+          biggest_os_flavor.set_hourly_cost_for(Atmosphere::OSFamily.first, 100)
+          biggest_os_flavor.reload
+          optimal_flavor.set_hourly_cost_for(Atmosphere::OSFamily.first, biggest_os_flavor.get_hourly_cost_for(Atmosphere::OSFamily.first))
```
mkasztelnik: Can you set here simply 100 instead of long getter? IMHO it is more readable.

spec/services/atmosphere/optimizer_spec.rb
```diff
@@ -555,6 +563,7 @@
           end
           @appl.dev_mode_property_set.preference_cpu = 4
 
+
```
mkasztelnik: Extra empty line

spec/services/atmosphere/optimizer_spec.rb
```diff
@@ -409,19 +409,27 @@
...
         end
 
         it 'selects flavor with more ram if prices are equal' do
           biggest_os_flavor = openstack.virtual_machine_flavors.max_by {|f| f.memory}
-          optimal_flavor = create(:virtual_machine_flavor, memory: biggest_os_flavor.memory + 256, cpu: biggest_os_flavor.cpu, hdd: biggest_os_flavor.hdd, hourly_cost: biggest_os_flavor.hourly_cost, compute_site: amazon)
+          optimal_flavor = create(:virtual_machine_flavor, memory: biggest_os_flavor.memory + 256, cpu: biggest_os_flavor.cpu, hdd: biggest_os_flavor.hdd, compute_site: amazon)
+          biggest_os_flavor.set_hourly_cost_for(Atmosphere::OSFamily.first, 100)
+          biggest_os_flavor.reload
+          optimal_flavor.set_hourly_cost_for(Atmosphere::OSFamily.first, biggest_os_flavor.get_hourly_cost_for(Atmosphere::OSFamily.first))
+          optimal_flavor.reload
```
mkasztelnik: why flavors need to be reloaded?

mkasztelnik:
I just finished first code review iteration. The biggest problem I see in billing service is the usage of static methods everywhere and violation of [Simple Responsibility Principle](http://en.wikipedia.org/wiki/Single_responsibility_principle). In the feature I propose to split this service into smaller classes, which can be instantiated and which can have local context (instead of passing state using method parameters).

spec/services/atmosphere/optimizer_spec.rb
nowakowski: Weird behavior of factories (possibly fixable by adding `autosave` to ActiveRecords; not tested).

spec/services/atmosphere/billing_service_spec.rb
nowakowski: Factorygirl weirdness. I'll try to fix it by updating the model class.

spec/services/atmosphere/billing_service_spec.rb
nowakowski: @mkasztelnik This will drastically increase the amount of code; moreover 3a and 3b share the same VMs - splitting them into two specs will be awkward. If the goal is to make specs more readable then I respectfully disagree with the proposed remedy.

spec/services/atmosphere/billing_service_spec.rb
nowakowski: This is done to force an invalid active record in order to test out the exception mechanism. I agree it's not a particularly useful spec.

spec/providers/atmosphere/optimization_strategy/manual_spec.rb
nowakowski: The spec finds both flavors valid for at (even though they differ w.r.t. CPU and RAM), so it selects the cheaper flavor - which is nondeterministic because the flavor factory produces flavors with random cost unless cost is manually set (as done here). Omitting these lines causes random Fs.

db/migrate/20150113100404_add_osfamilies.rb
nowakowski: Yeah, this is a leftover. Sorry; I forgot to clean up this comment.

app/services/atmosphere/billing_service.rb
nowakowski: Technically possible, because `atmosphere_deployments` has no primary key. The former method is safer. Thanks for the tip.

app/services/atmosphere/billing_service.rb
nowakowski: Hmm, it's a subjective opinion. I find it easier to analyze code when I don't have to hop around the tree trying to figure out what each method does. I agree that very complex conditions merit externalization but I don't think we've crossed this line here. :)

app/services/atmosphere/billing_service.rb
nowakowski: But that only reduces the call stack from 4 levels to 3. In order to satisfy the law, I would need to add `dep.os_family`, right? :)

app/providers/atmosphere/optimization_strategy/default.rb
nowakowski: Are you suggesting something like:
```
def get_hourly_cost_for(object)
  if object.is_a?(Atmosphere::VirtualMachineTemplate)
    # do something
  elsif object.is_a?(Atmosphere::OSFamily)
    # do something else
  else
    # exception
  end
end
```
...?

app/models/atmosphere/deployment.rb
nowakowski: No because I only want to run initial billing if the create process returns success - otherwise the appliance creation process should never proceed to the billing stage.

spec/services/atmosphere/billing_service_spec.rb
nowakowski: Yup, will fix.

spec/factories/appliance_types.rb
nowakowski: seed is a good idea; guarding against removal of last record is super paranoid - and the same logic applies to `Atmosphere::ComputeSite` and probably several other classes I can't think of right now. ;)

nowakowski:
I'll need a full day to iterate to 2nd version (Monday). I suggest we do not make this mess of a pull request any bigger by trying to cram in a rework of the billing service along the lines proposed by @mkasztelnik (splitting service into smaller classes). I will be happy to do it in a separate branch once we get this merged & deployed; agreed?

spec/serializers/atmosphere/virtual_machine_flavor_serializer_spec.rb
nowakowski: I guess I ran out of inspiration here. ;)

app/models/atmosphere/deployment.rb
mkasztelnik: But where/when these changes (`prepaid_until`, and `billing_state`) are saved?

nowakowski: OK, I think I understand what you mean. Poss. a bug.

app/providers/atmosphere/optimization_strategy/default.rb
mkasztelnik: Nop, it would be very ugly. You can reverse order here, e.g. `tmpl.hourly_cost(flavor)`.

mkasztelnik:
@nowakowski definitely not in the scope of this PR, it should be improved and merged ASAP.

app/models/atmosphere/virtual_machine_flavor.rb
tbartynski: @mkasztelnik Good point!

app/services/atmosphere/billing_service.rb
tbartynski: I see. Thanks!

db/migrate/20150103133015_move_prepaid_until_to_deployments.rb
nowakowski: It's a migration. It will only ever be run once on production data. :)

nowakowski: (I'll fix it ofc. :)

app/services/atmosphere/billing_service.rb
nowakowski: I ended up introducing `dep.os_family` instead.

app/services/atmosphere/billing_service.rb
nowakowski: Good catch. I have modified the code to run appl.save and dep.save transactionally, and to update the log outside of the transaction, ensuring that (1) consistency is maintained; (2) log message is written regardless of outcome.

spec/providers/atmosphere/optimization_strategy/manual_spec.rb
nowakowski: Suggestion: refactoring the spec is a self-contained subproject; let's move it to a separate branch/PR for the sake of expediency.

spec/requests/atmosphere/api/clew_spec.rb
nowakowski: To make `appl` "notice" the changes introduced by `initial_billing` in the appliance's deployment object. If you know a better way to fix it then I'm open to suggestions - moving `initial_billing` from `after_create` to `after_build` still does not update the in-memory state of `appl`. :(

spec/services/atmosphere/billing_service_spec.rb
nowakowski: We had a discussion regarding that once - the argument (which I concurred with) is that the `expect` condition should not duplicate values which are already set in the flavor factory - this introduces needless redundancy (updating the factory would also require updating the rspec accordingly). I'm hesitant to introduce this change although if you feel strongly about it then I will do so.

spec/services/atmosphere/billing_service_spec.rb
nowakowski: OK, I think this can be resolved by adding the obscure `inverse_of:` flag to the ActiveRecord relation. This establishes an identity-based link between A and B so that when A "has many" Bs and B is updated, the `A.Bs` reference will point to the updated object instead of the in-memory (outdated) version of A.Bs.

Added experimentally to `VirtualMachineFlavor` and 'VirtualMachineFlavorOSFamily` but I think this could help in many other places too.

spec/services/atmosphere/billing_service_spec.rb
nowakowski: This one tests regular billing (just to make sure everything is set up correctly). No harm in leaving it in, I think. The actual test comes later.

spec/services/atmosphere/optimizer_spec.rb
nowakowski: Seems `inverse_of:` in ActiveRecord relation fixes this problem.

nowakowski:
@mkasztelnik @tbartynski I've implemented your suggestions except as concerns refactoring specs - I believe it's wiser to do that in a separate pull request. I've also removed some excessively long lines but I generally disagree with Hound here - in this age of ultra-HD displays the 80 character limit is a classic http://www.retrologic.com/jargon/F/fossil.html :)

mkasztelnik:
Great, I will do second code review round on Monday. Can you also make this PR mergable with latest `master`? 

Also please take a look at #99 it is ready to be merged and it has already one thumb up :)

I cannot agree with your disagree :) No everyone is using whole HD display for only one window. For example take a look at my editor while working with atmosphere:

editor

So more/less 80 to 100 chars per line in every window opened (horizontal ruler is set to 80 chars). This limit also improve readability - you don't have to move your head to read whole line :)

nowakowski:
@mkasztelnik I've reviewed #99 as requested; found some minor issues which I hope @bwilk can address without much trouble.
Not sure why this isn't automatically mergeable but I'll perform the merge regardless after you give me those sweet :+1:s. ;)
Your screenshots indicate that your editor supports soft wrapping so where's the problem? ;)

mkasztelnik:
I have it turned on just to be able to read long lines without scrolling editor, but it is much more easy to read good formatted code that long single line - just saying, I'm not forcing you to change it if you don't want to :smile:

nowakowski:
I guess I can live with 100 characters per line... Hound still won't like me but that dog barks too much anyway. ;)

app/models/atmosphere/virtual_machine_flavor.rb
```diff
@@ -60,6 +66,37 @@ class VirtualMachineFlavor < ActiveRecord::Base
...
+      incarnation && incarnation.hourly_cost # nil if incarnation.blank?
+    end
+
+    # Upserts a binding between this VMF and an OS family, setting hourly cost
+    def set_hourly_cost_for(os_family, cost)
+      incarnation = virtual_machine_flavor_os_families.find_by(os_family: os_family)
+      if incarnation.blank?
+        incarnation = VirtualMachineFlavorOSFamily.new
+        incarnation.os_family = os_family
+        incarnation.virtual_machine_flavor = self
```
mkasztelnik: Please use reference for creation or object initialization.

```ruby
virtual_machine_flavor_os_families.initialize_by(os_family: os_family)
```

When you are using this method than `self` is aware about new object in relation otherwise (like in your current version) it doesn't.

mkasztelnik: It can also be simplified as follow:

```ruby
incarnation = virtual_machine_flavor_os_families.
  find_or_initialize_by(os_family: os_family)
incarnation.hourly_cost = cost
incarnation.save
```

app/serializers/atmosphere/virtual_machine_flavor_serializer.rb
```diff
@@ -5,7 +5,19 @@ class VirtualMachineFlavorSerializer < ActiveModel::Serializer
...
       :compute_site_id, :id_at_site,
-      :supported_architectures, :active
+      :supported_architectures, :active, :cost_map
+
+    private
+
+    # Deprecated: returns max hourly cost
+    def hourly_cost
+      object.virtual_machine_flavor_os_families.max_by(&:hourly_cost).hourly_cost
+    end
```
mkasztelnik: You don't need this method here, since it is already defined in `VirtualMachineFlavor`

app/serializers/atmosphere/virtual_machine_flavor_serializer.rb
```diff
@@ -5,7 +5,19 @@ class VirtualMachineFlavorSerializer < ActiveModel::Serializer
...
+
+    # Returns a full cost map for this flavor (depending on os_family)
+    def cost_map
+      object.cost_map
+    end
```
mkasztelnik: Similar situation as above. You don't need to define this method here. If it exists in serialized object it will be used without a problem.

app/services/atmosphere/billing_service.rb
```diff
@@ -20,45 +20,65 @@ def self.bill_all_appliances
...
-        raise Atmosphere::BillingException.new(message: "Unable to determine current payment validity date for appliance #{appl.id}")
+      appl.deployments.each do |dep|
+        bill_deployment(dep, billing_time, message, apply_prepayment)
+      end
+    end
+
+
+    def self.bill_deployment(dep, billing_time, message, apply_prepayment = true)
+      appl = dep.appliance # Shorthand
+      if dep.prepaid_until.blank?
```
mkasztelnik: It is not for now, but future PR: this condition smell to me like model validation. Maybe it can be moved there? I create dedicated issue to track this issue: #106

app/services/atmosphere/billing_service.rb
```diff
@@ -20,45 +20,57 @@ def self.bill_all_appliances
...
+        raise Atmosphere::BillingException.new(message: "Unable to determine current payment validity date for deployment #{dep.id} belonging to appliance #{appl.id}")
       else
-        # Figure out how long this appliance can continue to run without being billed again.
-        billing_interval = (billing_time - appl.prepaid_until) # This will return time in seconds
+        # Figure out how long this deployment can continue to run without being billed again.
+        billing_interval = (billing_time - dep.prepaid_until) # This will return time in seconds
         if billing_interval < 0
           # The appliance is still prepaid - nothing to be done.
-          Rails.logger.debug("Appliance #{appl.id} is prepaid until #{appl.prepaid_until}. Nothing to be done.")
+          Rails.logger.debug("Deployment #{dep.id} belonging to appliance #{appl.id} is prepaid until #{dep.prepaid_until}. Nothing to be done.")
```
mkasztelnik: For debug log with string concatenation please use following structure:

```
Rails.logger.debug { "this is #{my} debug message" }
```

of (no newlines thanks to `squish`)

```ruby
Rails.logger.debug do
  <<-LOG.squish
  My long
  debug message
  with #{variables}
  LOG
end
```

the difference is that example presented above will evaluate lambda only when debug is turned on. In original code string in interpolated every time, even if debug is turned off.

app/services/atmosphere/billing_service.rb
```diff
@@ -20,45 +20,65 @@ def self.bill_all_appliances
...
               appl.last_billing = billing_time
-              appl.billing_state = "prepaid"
-              appl.save
-              if appl.errors.blank?
+              dep.prepaid_until = billing_time+(apply_prepayment ? @appliance_prepayment_interval : 0)
+              dep.billing_state = "prepaid"
+              # These should be saved together or not at all
+              ActiveRecord::Base.transaction do
+                appl.save
+                dep.save
```
mkasztelnik: `appl.fund.save` is invoked outside transaction - is it ok?

app/services/atmosphere/billing_service.rb
```diff
@@ -72,48 +92,46 @@ def self.bill_appliance(appl, billing_time, message, apply_prepayment = true)
...
-    # This method calculates the amount which should be billed to a specific appliance on billing_time.
+    # This method calculates the amount which should be billed to a specific deployment on billing_time.
     # It does not actually incur the charge, merely returns the appropriate information.
     # billing_time specifies when the billing action is actually performed
     # apply_prepayment determines whether the method should include prepayment period (of length specified in @appliance_prepayment_interval)
-    # - This is usually true, but will be false when calculating final charge for an appliance which is being shut down.
-    def self.calculate_charge_for_appliance(appl, billing_time, apply_prepayment = true)
-      # By default, the billing process should extend the validity of the appliance until now+1.hour.
-      billable_time = self.calculate_billable_time(appl, billing_time, apply_prepayment)
+    # - This is usually true, but will be false when calculating final charge for a deployment which is being shut down.
```
mkasztelnik: This also can be done later: when documentation for the method is created can you use [Rdoc format](http://rdoc.sourceforge.net/doc/index.html)?

db/migrate/20150113100404_add_osfamilies.rb
```diff
@@ -0,0 +1,59 @@
+# This migration creates all the necessary schema models to ensure flavor and appliance type
+# differentiation into OS families (typically Linux and Windows). This is necessary to facilitate
+# proper billing where the price of a flavor depends on the installed OS.
+
+class AddOsfamilies < ActiveRecord::Migration
+  def up
+    create_table :atmosphere_os_families do |t|
+      t.string :os_family_name, null:false, default: "Windows"
```
mkasztelnik: Why so long name? Maybe `OsFamily.name` would be better?

db/migrate/20150113100404_add_osfamilies.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    create_table :atmosphere_virtual_machine_flavor_os_families do |t|
```
mkasztelnik: Maybe `FlavorOsFamily`? The name `VirtualMachineFlavorOsFamily` is soooooooo long and all relation names are sooooooo long.

spec/services/atmosphere/billing_service_spec.rb
mkasztelnik: I propose to remove it.

spec/services/atmosphere/billing_service_spec.rb
mkasztelnik: That is why it should be split into 2 specs - one testing regular billing, second final billing - for sure TODO for next PR.

app/services/atmosphere/billing_service.rb
nowakowski: It's not - good catch!

nowakowski:
@mkasztelnik Done. I've refactored class and attribute names so you'll need to rerun your migrations locally.

Since Travis said "no", I forced an update to rvm 2.1.4. I think it was due anyway.

nowakowski:
@tbartynski @Nuanda @mkasztelnik Guys...?

app/models/atmosphere/virtual_machine_flavor.rb
```diff
@@ -60,6 +66,33 @@ class VirtualMachineFlavor < ActiveRecord::Base
 
     scope :active, -> { where(active: true) }
 
+    # Assumes only 1 vmf_osf for a specific os_family
+    def get_hourly_cost_for(os_family)
+      incarnation = flavor_os_families.find_by(os_family: os_family)
+      incarnation && incarnation.hourly_cost # nil if incarnation.blank?
```
Nuanda: Is that really nil or simply false (the method call result) if incarnation.blank?

app/models/atmosphere/virtual_machine_template.rb
```diff
@@ -144,6 +144,12 @@ def perform_delete_in_cloud
       logger.warn("VMT with #{id_at_site} does not exist - continuing")
     end
 
+    # Returns the hourly cost for this template assuming a given VM flavor
+    def get_hourly_cost_for(flavor)
+      incarnation = flavor.flavor_os_families.find_by(os_family: appliance_type.os_family)
+      incarnation && incarnation.hourly_cost
+    end
```
Nuanda: Again this kind of execution. I'm not convinced it's going in the right direction. I understand the intent - make sure we do not fail with the infamous 'nil has no method called hourly_cost' exception but, what we get instead, is that the code that calls this method receives false (and it expected a number, I guess). So, we'll probably get an exception elsewhere anyway (it will be something like "cannot compare false to 12.4").

app/models/atmosphere/virtual_machine_flavor.rb
mkasztelnik: I is very simple checkable, just run this in pry or irb:

```ruby
a = nil
a && a[:a]

a = { a: 'a' }
a && a[:a
```

As you can see everything is as expected here :smile: 

```ruby
[18] pry(main)> a = nil
=> nil
[19] pry(main)> a && a[:a]
=> nil
[20] pry(main)> a = { a: 'a' }
=> {:a=>"a"}
[21] pry(main)> a && a[:a]
=> "a"
```

app/models/atmosphere/virtual_machine_template.rb
mkasztelnik: See my previous comment - everything is ok here, `nil` will be returned.

Nuanda:
Well, it's a huge push with a lot of changes, discussions, fixes... I don't think I can any longer propose any meaningful comments to this except nitpicking. Sure some things could be written a bit different but perhaps, for the sake of the project, should we move on?

mkasztelnik:
@nowakowski I will do third round (hopefully the last one :smile:) tomorrow after imsop meeting.

app/models/atmosphere/virtual_machine_template.rb
Nuanda: How returning nil in a method that clearly intents to return a number is "ok"?

mkasztelnik: I though that you were referring to the `something && something.property` structure in first comment :smile: You are absolutely right that `nil` should not be returned here but maybe default value.

@nowakowski what is your opinion?

nowakowski: @mkasztelnik I have given this quite a lot of thought and IMHO no sensible "default value" can be returned in this context. Think about it: we are asking for the price of something which does not exist. What should the answer be? 0? 65535? +infinity? ;) Worse yet, if we return a numerical value, the client will later expect this exact cost to be applied when billing the appliance, complicating our model and also playing havoc with optimizer logic:

- How much does this cost?
- $1000.
- Okay, I want to buy it!
- Uh... well, you see... perhaps I could interest you in this other thing instead...? ;)

Realistically speaking, we have a choice between:
1. returning nil (which is what currently happens)
2. throwing an exception (which is a reasonable alternative)

In both cases the ball is in the client's court. If you feel strongly about choosing (2) over (1) then I will change the method accordingly.

mkasztelnik: @nowakowski if result of this function is only presented to the user (not used by other Atmosphere mechanisms) than I'm ok with `nil` value.

nowakowski: @mkasztelnik It is used by Atmosphere mechanisms.

mkasztelnik: I see... It is used in `Atmosphere::OptimizationStrategy::Default#select` - so as @Nuanda correctly mentioned nil will lead to `NoMethodError: undefined method `<' for nil:NilClass` exception. 

Of course this is a situation when we will not have appropriate `os_family` in our DB. I propose to merge it and deploy it with guaranteeing correct `os_families` in our DBs and submit issue describing this problem which will be fixed ASAP. What do you think?

mkasztelnik:
@nowakowski can you rebase into latest master and check if everything is working as expected? If everything is ok than at the beginning I propose to merge #114 since it solves quite serious bug and #99 (migrations for this PR are already on development), deploy it into production and then merge this one into master.

Are you OK with this plan?

nowakowski:
@mkasztelnik Yup, but I cannot do so before Monday (there's a 50% chance I might find some time to review your latest comments on Sunday).

app/models/atmosphere/virtual_machine_template.rb
nowakowski: @mkasztelnik  @Nuanda I'm fine with fixing this problem right here and now but you need to tell me what you want me to do. Generally speaking, the situation we're dealing with here is a "can't happen" error, i.e. I do not expect `f.get_hourly_cost_for` to ever be invoked on an `OSFamily` object which is not already bound to `f`. This whole "guard against nils" code is just my paranoia - I want to be a good boy scout and not blindly invoke a custom method on an object which is passed to my code - simply because I do not control the requestor and cannot guarantee that the input argument is not blank. As already suggested, I can either respond with nil ("garbage in, garbage out") or toss an exception - whichever you prefer.

Nuanda: It's fine with me the way it is. If to be changed, I would vote for neither solution. In fact, I would simply write it as "incarnation.hourly_cost" and that's it. Exporting "nil exception" higher in the stack does nothing except making the bug search time longer. Raising exception - this is only good as long as you have a serious plan what to do with it. If not (i.e. no appropriate code somewhere higher up, to catch it and act accordingly) this is just changing "nil exception" into "my exception". So -> not worth it.

Now I know this "a && a.call" construct is prevalent these days in open source projects. Perhaps there are places this makes more sense (i.e. when the method is supposed to return an AR object, like a scope or a specific find method, and then receiving nil is to be expected and properly dealt with on the "other side"). In most of the cases, though, in my opinion, it just makes the code more cryptic, longer, and does not "save the day" at all when something wrong happens.

So, @nowakowski, just leave it as it is :).

nowakowski: @Nuanda I'm inclined to agree with this logic.

mkasztelnik: The is another possibility. While destroying `OsFamily` we can check if there is any template with this family assigned. If yes then refuse to destroy it.

nowakowski: @mkasztelnik This won't address the issue at hand (a template may represent an OS family which does not exist on its current compute site - quite likely, in fact, if a Windows template is copied to a Linux-only site). Besides, I can't imagine a situation where an `OSFamily` object is destroyed - except perhaps when added in error.

nowakowski:
Travis is hiccuping. Excuse my mess.

nowakowski:
@bwilk I think your removal of `schema.rb` is causing Travis to crap out because it uses it to bootstrap the test DB. Just a hypothesis. Could you comment on why this file was removed?

nowakowski:
Ok, fixed - but I had to recommit `schema.rb`.

PR #71, opened 2015-01-05
remove unused private method

mkasztelnik:
It seems like private `save_template_in_cloud` method is not used - removing.


mkasztelnik:
@nowakowski, @Nuanda please take a look

Nuanda:
There is only this single occurrence of this string in the project, so given it's a private method, this should be safe.
:+1:

paoolo:
I would say the same what @Nuanda and :+1:

PR #72, opened 2015-01-05
auditable concern for user actions

mkasztelnik:
We are repeating all over again the same schema for logging user actions. Auditable concern is trying to encapsulate this feature in one place.

/cc @Nuanda, @nowakowski please take a look


nowakowski:
:+1:

PR #73, opened 2015-01-07
add missing asert for forbidden post action

mkasztelnik:
in 

```ruby
it 'returns 403 Forbidden when creating configuration template for not owned appliance type' do
  post api("/appliance_configuration_templates", user), new_not_owned_config_template_request
end
```

following line was missing

```ruby
expect(response.status).to eq 403
```

/cc @dice-cyfronet/atmo-dev please take a look


paoolo:
:+1:

Nuanda:
:+1:

PR #77, opened 2015-01-12
user details for current_user

### LINKED ISSUE #76: User roles in /user REST API for current user
Right now current user roles are not returned in the `/users` REST API. As a consequence `clew` is using   hacks (specific for concrete deployment) to get information if user is an admin or developer. To correct this situation and create generic solution we should add `roles` element into `UserSerializer`. `roles` element should be filled only for `current_user`

/cc @dharezlak

mkasztelnik:
Fixes #76 

/cc @dice-cyfronet/atmo-dev-team, @dharezlak please take a look

### DIFF
--- a/app/serializers/atmosphere/user_serializer.rb
+++ b/app/serializers/atmosphere/user_serializer.rb
@@ -9,11 +9,17 @@ module Atmosphere
 
     def attributes
       hash = super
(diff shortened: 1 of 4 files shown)


tbartynski:
:+1:

nowakowski:
:+1:

PR #83, opened 2015-01-13
Ensure default fund

### LINKED ISSUE #81: When first fund is assigned to the user it should be marked as default

Nuanda:
Fixes #81

### DIFF
--- a/app/models/atmosphere/user_fund.rb
+++ b/app/models/atmosphere/user_fund.rb
@@ -22,5 +22,37 @@ module Atmosphere
                   scope: :fund_id,
                   message: I18n.t('funds.unique_user')
               }
+
+    around_save :ensure_single_default
+    after_destroy :reallocate_default
+
+
+    private
+
+    def ensure_single_default
+      unless self.default
+        ufs = UserFund.where(user_id: user_id)
+        if ufs.count == 0 || (ufs.count == 1 && ufs[0].id == self.id)
+          self.default = true
+        end
+      end
+
+      yield
+
+      if self.default
+        ufs = UserFund.where(user_id: user_id, default: true)
+        (ufs - [self]).each do |uf|
+          uf.update_attributes(default: false)
+        end
+      end
+    end
+
+    def reallocate_default
+      ufs = UserFund.where(user_id: user_id)
+      if ufs.count > 0 && ufs.none?(&:default)
+        ufs.first.update_attributes(default: true)
+      end
+    end
+
   end
 end
--- a/app/models/atmosphere/user_fund.rb
+++ b/app/models/atmosphere/user_fund.rb
@@ -24,7 +24,7 @@ module Atmosphere
               }
 
     around_save :ensure_single_default
-    after_destroy :reallocate_default
+    after_destroy :reallocate_default if :default
 
 
     private
@@ -48,11 +48,8 @@ module Atmosphere
     end
 
     def reallocate_default
-      ufs = UserFund.where(user_id: user_id)
-      if ufs.count > 0 && ufs.none?(&:default)
-        ufs.first.update_attributes(default: true)
-      end
+      uf = UserFund.find_by(user_id: user_id)
+      uf && uf.update_attributes(default: true)
     end
-
   end
 end
--- a/app/views/atmosphere/admin/funds/_user_fund.html.haml
+++ b/app/views/atmosphere/admin/funds/_user_fund.html.haml
@@ -1,0 +1,36 @@
+.btn-group
+  - user = user_fund.user
+
+  %span.btn.btn-default.btn-xs.disabled.non-opaque
+    = user.full_name
+    = surround '(', ')' do
+      = user.login
+
+  - if user_fund.default
+    %span{ class: 'btn btn-xs btn-primary has_bottom_tooltip',
+    title: t('funds.set_default.to_false',
+             user: user_fund.user.full_name) }
+      D
+  - else
+    .btn-group
+      = form_for [:admin, user_fund] do |f|
+        = f.hidden_field :default, value: true
+        %button.btn.btn-xs.btn-default.has_bottom_tooltip{ type: 'submit',
+        title: t('funds.set_default.to_true',
+                 fund: user_fund.fund.name,
+                 user: user_fund.user.full_name) }
+          D
+
+  = link_to admin_billing_logs_path(user_id: user.id),
+    remote: true,
+    data: { billing: 'on' },
+    title: t('billing.load_for_user', user: user.full_name),
+    class: 'btn btn-default btn-xs has_bottom_tooltip' do
+    = icon 'bar-chart'
+
+  = link_to admin_user_fund_path(user_fund),
+    method: :delete,
+    data: { confirm: t('are_you_sure') },
+    class: 'btn btn-default btn-xs has_bottom_tooltip',
+    title: t('funds.remove_user.help', user: user_fund.user) do
+    = icon 'times'
--- a/spec/models/atmosphere/user_fund_spec.rb
+++ b/spec/models/atmosphere/user_fund_spec.rb
@@ -61,18 +61,17 @@ describe Atmosphere::UserFund do
     end
 
     it 'does not prematurely default to an invalid fund' do
-      begin
-        create(:user_fund,
-               fund: user_fund.fund,
-               user: user_fund.user,
-               default: true)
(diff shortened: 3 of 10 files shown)


app/models/atmosphere/user_fund.rb
```diff
@@ -22,5 +22,37 @@ class UserFund < ActiveRecord::Base
...
+        end
+      end
+    end
+
+    def reallocate_default
+      ufs = UserFund.where(user_id: user_id)
+      if ufs.count > 0 && ufs.none?(&:default)
+        ufs.first.update_attributes(default: true)
+      end
+    end
```
mkasztelnik: This method can be simplified into:

```ruby
def reallocate_default
  uf = UserFund.find_by(user_id: user_id)
  uf.update_attributes(default: true) if default && uf
end
```

app/views/atmosphere/admin/funds/_user_fund.html.haml
```diff
@@ -0,0 +1,36 @@
...
+      = user.login
+
+  - if user_fund.default
+    %span{ class: 'btn btn-xs btn-primary has_bottom_tooltip',
+    title: t('funds.set_default.to_false',
+             user: user_fund.user.full_name) }
+      D
+  - else
+    .btn-group
+      = form_for [:admin, user_fund] do |f|
```
mkasztelnik: This have one small drawback - whole webpage is refreshed and selection for fund graph disappeared. This is very, very minor issue and if complicated don't have to be changed :smile:

spec/models/atmosphere/user_fund_spec.rb
```diff
@@ -33,4 +33,75 @@
...
+
+    it 'makes first user_fund default' do
+      expect(user_fund.default).to be_truthy
+    end
+
+    it 'blocks setting the only fund as not default' do
+      expect(user_fund.default).to be_truthy
+      user_fund.default = false
+      user_fund.save
+      expect(user_fund.reload.default).to be_truthy
```
mkasztelnik: Maybe instead of reloading the object you can check for `user_fund` errors?

spec/models/atmosphere/user_fund_spec.rb
```diff
@@ -33,4 +33,75 @@
...
+
+    it 'does not bother single default fund creation' do
+      pc = create(:poor_chap)
+      uf = create(:user_fund, user: pc, default: true)
+      expect(uf.reload.default).to be_truthy
+    end
+
+    it 'keeps only single default user fund' do
+      uf = create(:user_fund, user: user_fund.user, default: true)
+      expect(Atmosphere::UserFund.count).to eq 2
```
mkasztelnik: Why this is needed? Here you know exactly that `uf` and `user_fund` are in the database.

spec/models/atmosphere/user_fund_spec.rb
```diff
@@ -33,4 +33,75 @@
...
+      expect(user_fund.reload.default).to be_falsey
+    end
+
+    it 'does not prematurely default to an invalid fund' do
+      begin
+        create(:user_fund,
+               fund: user_fund.fund,
+               user: user_fund.user,
+               default: true)
+      rescue ActiveRecord::RecordInvalid
```
mkasztelnik: This can be changed into following construction:

```ruby
expect do
  create(:user_fund, ...)
end.to raise_error(ActiveRecord::RecordInvalid)
```

or (maybe even better)

```ruby
expect do
  create(:user_fund, ...)
end.to change { Atmosphere::UserFund.count }.by 0
```

spec/models/atmosphere/user_fund_spec.rb
```diff
@@ -33,4 +33,75 @@
...
+      expect(user_fund.reload.default).to be_truthy
+    end
+  end
+
+
+  describe '#reallocate_default' do
+
+    it 'does nothing when last fund is removed' do
+      user_fund.destroy
+      expect(Atmosphere::UserFund.count).to eq 0
```
mkasztelnik: I'm not sure if I understand the purpose of this test.

spec/models/atmosphere/user_fund_spec.rb
```diff
@@ -33,4 +33,75 @@
...
+    end
+
+    it 'reallocates removed default' do
+      uf = create(:user_fund, user: user_fund.user)
+      uf2 = create(:user_fund, user: user_fund.user)
```
mkasztelnik: The is already spec checking if only one user fund can be set to default. As a conclusion maybe only one additional user fund (`uf`) can be created here (and the same situation in spec below) to check reallocation conditions?

mkasztelnik:
I have only very small remarks/propositions for improvements, but the only thing I can write here is :+1:

 :clap: :clap:

app/models/atmosphere/user_fund.rb
mkasztelnik: One additional improvement

```ruby
after_destroy :reallocate_default if :default

def reallocate_default
  uf = UserFund.find_by(user_id: user_id)
  uf && uf.update_attributes(default: true)
end
```

Nuanda: all right

app/views/atmosphere/admin/funds/_user_fund.html.haml
Nuanda: Yes, I know. Alternatives are quite complicated imho though.

spec/models/atmosphere/user_fund_spec.rb
Nuanda: There will be no errors since this is not regarding validation. Without reload most of those tests gave me false positives or false negatives. So I decided to be on the safe side so I reload them.

spec/models/atmosphere/user_fund_spec.rb
Nuanda: Sure. Unless the user factory (and you need a user there) creates two different funds just for the fun of it! So I end up spending a better part of an hour figuring these things out. Then I switched to my better :poor_chap factory but I decided to those leave "theoretically stupid" (but, in fact, quite useful) expectations in the code.

spec/models/atmosphere/user_fund_spec.rb
Nuanda: Yes, but then @mkasztelnik will tell me this spec tests for two different things and it is not acceptable by our standards ;).

spec/models/atmosphere/user_fund_spec.rb
Nuanda: Very useful. Tests for boundary cases - something we should test more. Perhaps, actually, those are the only spec tests worth doing (since many errors are because, for instance, callbacks assuming there is at least one object in the DB and failing with exceptions where there are none). In such a case this test should end with an exception and that will be easier to notice by a developer.

spec/models/atmosphere/user_fund_spec.rb
Nuanda: No, since we have different callbacks called here. So no - this was not tested before (saving/creating/updating was).

Nuanda:
Marek, If you want to commit the changes regarding the callback syntax, and all tests pass, then go ahead.

spec/models/atmosphere/user_fund_spec.rb
mkasztelnik: Ok I understand. I also discovered this issue that some of the factories do to much. We need to take a look at them in the feature

spec/models/atmosphere/user_fund_spec.rb
mkasztelnik: Eeee, I'm not that bad :smiling_imp:

Nuanda: Just kidding. Go ahead and add the raise expectation, if you have a second.

Nuanda:
I would ignore the self.default complaints of the houndini this time. First of all, they are acutally not all redundant, I think. And single 'default' is such a generic name, I decided to place the exact self. reference (what I usually _not_ do) to make it easier to read and understand.

mkasztelnik:
@Nuanda as far as I know `self` is required only when you are setting value, for getters for sure `self` is redundant - but I understand your rationale :smile: 

:+1:

Nuanda:
Thanks for commits.

Nuanda:
@dice-cyfronet/atmo-dev-team another code review please

nowakowski:
:+1:

PR #93, opened 2015-01-20
Added a more verbose warning message when deleting CS via the admin GUI.

### LINKED ISSUE #90: Inform user about destructive operation side-effects
After removing Compute Site from Atmosphere, VMTs located on that Compute Site are removed. I guess that flag `managed_by_atmosphere` decided if the VMT will be removed or not. But there is no information displayed for admin about side-effects of removing Compute Site.

/cc @nowakowski @mkasztelnik @Nuanda what are you thinking about informing admin about side-effects of some operation performed via Atmosphere?

nowakowski:
@mkasztelnik @paoolo Please review.

Fixes #90

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -37,7 +37,7 @@ en:
   keep_it_secret: Keep it secret, keep it safe!
   reset: Reset
   are_you_sure: Are you sure?
-  are_you_sure_cs: Warning! This will delete all Virtual Machine Templates which exist on this CS. Are you sure?
+  are_you_sure_cs: Warning! This will irrevocably destroy all Virtual Machine Templates and Virtual Machines running on this CS which have been created by Atmosphere. Are you sure?
   actions: Actions
   add_new: Add new
   editing: Editing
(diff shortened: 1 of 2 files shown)


mkasztelnik:
@nowakowski should I tag it as ready for review?

config/locales/en.yml
```diff
@@ -37,6 +37,7 @@ en:
   keep_it_secret: Keep it secret, keep it safe!
   reset: Reset
   are_you_sure: Are you sure?
+  are_you_sure_cs: Warning! This will delete all Virtual Machine Templates which exist on this CS. Are you sure?
```
paoolo: May you say "which exist on this CS and are managed by Atmosphere"? But I guess that safer is this what you proposed.

mkasztelnik:
IMHO this is not enough - also virtual machines and flavors will be deleted. What is more note about managed by atmosphere would also be useful.

nowakowski:
@mkasztelnik Sorry, I forgot about the tags.

nowakowski:
@mkasztelnik There you go.

mkasztelnik:
:+1:

Nuanda:
:+1:

Issue #86, opened 2015-01-15
Disable Hound CI checking changes in `schema.rb`


paoolo:
For files:

* `db/schema.rb`
* `spec/dummy/db/schema.r`

See: https://github.com/thoughtbot/hound/issues/480

@Nuanda, @mkasztelnik  What are you thinking about this?

mkasztelnik:
I think that exclude for `schema.rb` is a good solution - go for it :sunglasses:. Also configuration of `hound-tools` looks nice - can you try to configure atmosphere to use it?

paoolo:
I will try it.

paoolo:
PR #94 created. How to check this how it works if Hound verify only committed changes, not whole project?

paoolo:
Closed via #94

PR #94, opened 2015-01-22
Disable `schema.db` from Hound review

paoolo:
Due to #86

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,4 +1,3 @@
-# Prefer single quotes
-StringLiterals:
-  EnforcedStyle: single_quotes
-  Enabled: true
+ruby:
+  enabled: true
+  config_file: .rubocop.yml
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,11 @@
+AllCops:
+  RunRailsCops: true
+  Exclude:
+    - db/schema.rb
+    - spec/dummy/db/schema.rb
+    - Gemfile
+    - Rakefile
+
+StringLiterals:
+  EnforcedStyle: single_quotes
+  Enabled: true
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,7 +1,6 @@
 AllCops:
   RunRailsCops: true
   Exclude:
-    - db/schema.rb
     - spec/dummy/db/schema.rb
     - Gemfile
     - Rakefile


mkasztelnik:
Hound defaults are needed to run check locally, right?

paoolo:
@mkasztelnik I don't think so. I did this what was describe in https://github.com/e2/hound-tools/blob/master/README.md

I ran `bundle exec hound-tools init` and commit changes.

paoolo:
I will test how changed Hound settings affect Hound code review in repositories https://github.com/dice-cyfronet/migratio and https://github.com/dice-cyfronet/ip-wrangler

.hound.yml
```diff
@@ -1,4 +1,3 @@
-# Prefer single quotes
-StringLiterals:
-  EnforcedStyle: single_quotes
-  Enabled: true
```
mkasztelnik: Why did you remove these lines?

mkasztelnik: Most probably is should be added into `.hound/overrides.yml` and regenerated.

paoolo: Added in 78a9be2

mkasztelnik: Hmm I see. So now it is added into rubocop defaults but not into hound customization. Did you removed lines above manually or it was done when `bundle exec hound-tools` was executed?

paoolo: I moved it manually.

mkasztelnik: As far as I understand `hound-tools` documentation it does not touch ``hound.yml`, so these changes should be there. What is more information about excluding `schema.rb` should also be added here. 

I don't understand why `hound-tools` cannot take overrides from `.hound.yml` but take them from `.hound/overrides.yml`.

mkasztelnik:
@paoolo I was looking into other projects which use `hound-tools`. It seems like it is not so popular... What is more in documentation I see different files that committed here.

paoolo:
OK. I will check if manual updates in rubocop and hound config will give 
the same result.

Nuanda:
If you intend to add more commits to a branch, remove the ready-for-review label from the PR.

.hound.yml
paoolo: I disabled `hound-tools`. I moved this lines to `.rubocop.yml`

mkasztelnik: Ok, but be sure to check if `.rubocop.yml` is loaded by hound.

paoolo: Can I use this changes in [`feature_migrating_virtual_machine_templates`](https://github.com/dice-cyfronet/atmosphere/tree/feature_migrating_virtual_machine_templates)?

paoolo: Testing in #103

paoolo:
@dice-cyfronet/atmo-dev-team updated configuration. Does it fit you?

.rubocop.yml
```diff
@@ -0,0 +1,11 @@
+AllCops:
+  RunRailsCops: true
+  Exclude:
+    - db/schema.rb
```
mkasztelnik: You can remove this line since @bwilk find that `db/schema.rb` is relevant and it will be remove when #99 is merged.

paoolo: Why `db/schema.rb` will be removed from repo?

mkasztelnik:
One small remark reported above, when it is done :+1: from my side :smiley:

.rubocop.yml
mkasztelnik: It is simply not needed. While updating db in `dummy` project from the spec `spec/dummy/db/schema.rb` is updated no `db/schema.rb`

paoolo: Done. Removed.

bwilk:
Second thumb up :+1:

PR #95, opened 2015-01-23
take into account created at when checking vm age

mkasztelnik:
Childhood age is used to fix the problem when following situation occurs:
  - atmosphere send request to start VM
  - after request is sent monitoring is triggered, new VM is discovered and registered as `managed_by_atmosphere = false`
  - atmosphere is trying to save VM created in first step into DB. As a result  exception is thrown (VM with id_at_site already exists)

We used `server.created` from compute site to fix this problem. It has one drawback - when compute site host data is different that date on the host where atmosphere is hosted. To minimize this problem this commit introduces checking at the beginning `created_at` field and only when it is empty `created` date returned by the compute site is used.

/cc @dice-cyfronet/atmo-dev-team please take a look.

### DIFF
--- a/app/services/atmosphere/vm_updater.rb
+++ b/app/services/atmosphere/vm_updater.rb
@@ -45,7 +45,7 @@ module Atmosphere
     end
 
     def old_enough?
-      server.created < Atmosphere.childhood_age.seconds.ago
+      (vm.created_at || server.created) < Atmosphere.childhood_age.seconds.ago
     end
 
     def furhter_update_requred?
--- a/spec/services/atmosphere/vm_updater_spec.rb
+++ b/spec/services/atmosphere/vm_updater_spec.rb
@@ -11,10 +11,10 @@ describe Atmosphere::VmUpdater do
   let(:updated_vm) { Atmosphere::VirtualMachine.find_by(id_at_site: 'id_at_site') }
 
   before do
-    allow(Atmosphere::VirtualMachineTemplate)
-      .to receive(:find_by)
-        .with(compute_site: cs, id_at_site: "vmt_id_at_site")
-          .and_return(vmt)
+    allow(Atmosphere::VirtualMachineTemplate).
+      to receive(:find_by).
+      with(compute_site: cs, id_at_site: 'vmt_id_at_site').
+      and_return(vmt)
     #Zabbix.stub(:register_host).and_return 1
     #Zabbix.stub(:unregister_host)
     #Zabbix.stub(:host_metrics)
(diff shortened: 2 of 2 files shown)


mkasztelnik:
@mpawlik problem with previous solution was discovered on ismop installation, where atmosphere hosting machine date time is misconfigured (more/less 2 minutes delay) - as a conclusion we needed to wait more than 2 minutes for status change. Now patched version is installed there and starting VM process is taking 5 to 10 seconds.

mkasztelnik:
@dice-cyfronet/atmo-dev-team anyone?

Nuanda:
Well, I checked this PR twice and I don't feel competent to review this. Leaving it for others.

mkasztelnik:
@tbartynski, @nowakowski waiting for your feedback?

spec/services/atmosphere/vm_updater_spec.rb
```diff
@@ -289,11 +308,11 @@
...
-          updated_at_site: nil,
-          id_at_site: 'id_at_site',
-          ip: nil,
-          compute_site: cs
-        )
+                  updated_at_site: nil,
+                  id_at_site: 'id_at_site',
+                  ip: nil,
+                  compute_site: cs,
+                  created_at: 10.seconds.ago)
```
nowakowski: Perhaps use a config variable instead of this hardcoded interval?

spec/services/atmosphere/vm_updater_spec.rb
```diff
@@ -304,11 +323,11 @@
...
-          updated_at_site: vm_update_at,
-          id_at_site: 'id_at_site',
-          ip: nil,
-          compute_site: cs
-        )
+                  updated_at_site: vm_update_at,
+                  id_at_site: 'id_at_site',
+                  ip: nil,
+                  compute_site: cs,
+                  created_at: 10.seconds.ago)
```
nowakowski: Same as above.

mkasztelnik:
@nowakowski changed as requested. @tbartynski + other @dice-cyfronet/atmo-dev-team member please, please, please take a look at this PR - it is 1 week old already and only response from @Nuanda and @nowakowski  :cry:

nowakowski:
:+1:

app/services/atmosphere/vm_updater.rb
```diff
@@ -45,7 +45,7 @@ def updated_at
     end
 
     def old_enough?
-      server.created < Atmosphere.childhood_age.seconds.ago
+      (vm.created_at || server.created) < Atmosphere.childhood_age.seconds.ago
```
paoolo: How this change affecting the functioning of VM discovery mechanism?

mkasztelnik: Discovery mechanism is not affected here. Please notice that `created_at` is `nil` for not new Active Record element - then `server.created` will be taken into account.

tbartynski:
:+1:

PR #96, opened 2015-01-27
200 and 204 as success while deleting dnat redirection

mkasztelnik:
Necessary changes required when dice-cyfronet/ip-wrangler#7 PR will be merged and deployed.

Introduced changes are backward compatible.

/cc @dice-cyfronet/atmo-dev-team please take a look


paoolo:
Response body will be examined? If DELETE return 200 it will return JSON table with hashes pointed to removed DNAT rules.

mkasztelnik:
Nop, we don't care :smile:

paoolo:
OK.

Nuanda:
:+1:

paoolo:
:+1:

PR #97, opened 2015-01-27
allow apps to use this engine's factories in rspecs

Nuanda:
I need to use factories from Atmo specs in PLGrid application. This fix is done according to this solution:

http://stackoverflow.com/questions/20261585/how-to-use-factorygirl-factories-from-an-engine


mkasztelnik:
:+1:

paoolo:
:+1:

PR #99, opened 2015-01-30
Action log

bwilk:
Action log implemented. Please review and merge. 
@mkasztelnik, @tbartynski


app/models/atmosphere/action.rb
```diff
@@ -0,0 +1,13 @@
+module Atmosphere
+  class Action < ActiveRecord::Base
```
mkasztelnik: In the feature `Action` will also be connected with ApplianceType - it will be added in another PR - am I right?

app/serializers/atmosphere/action_log_serializer.rb
```diff
@@ -0,0 +1,10 @@
+#
+# Action log serializer.
+#
+module Atmosphere
+  class ActionLogSerializer < ActiveModel::Serializer
+
+    attributes :id, :log_level, :message, :created_at
```
mkasztelnik: IMHO `appliance_id` is missing here.

app/services/atmosphere/optimizer.rb
```diff
@@ -14,6 +14,12 @@ def run(hint)
 
     #private
     def satisfy_appliance(appliance)
+
+      action = Action.create { |action|
+        action.appliance = appliance
+        action.action_type = :satisfy_appliance
+      }
```
mkasztelnik: It can be simplified using following construction: 

```ruby
Action.create(appliance: appliance, action_type: :satisfy_appliance)
```

mkasztelnik: Here and in scaling you are creating only `action`, no `action_logs` are added. Maybe information when scaling started/finished, what is the results (e.g. id of VMs started/reused) should be logged here.

The same comment is valid for scale method.

app/controllers/atmosphere/api/v1/actions_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+                                    class: 'Atmosphere::Action'
+
+        respond_to :json
+
+        def index
+          respond_with @actions.where(filter).order(:id).uniq
+        end
+
+        def show
+          respond_with @actions
```
tbartynski: By convention show should respond with @action not @actions.

app/services/atmosphere/optimizer.rb
```diff
@@ -80,6 +86,10 @@ def select_tmpls_and_flavors(tmpls, options={})
     def scale(hint)
       appliance = hint[:appliance]
       quantity = hint[:quantity]
+      action = Action.create { |action|
+        action.appliance = appliance
+        action.action_type = :scale
+      }
```
mkasztelnik: The same as above. Use `Action.create(attr: val)` syntax.

spec/factories/action_logs.rb
```diff
@@ -0,0 +1,6 @@
+FactoryGirl.define do
+  factory :action_log, :class => 'Atmosphere::ActionLog' do
```
mkasztelnik: Please use ruby 1.9 syntax (`class: 'Atmosphere::ActionLog'`)

spec/factories/actions.rb
```diff
@@ -0,0 +1,10 @@
+FactoryGirl.define do
+  factory :action, :class => 'Atmosphere::Action' do
```
mkasztelnik: The same as above - ruby 1.9 syntax

spec/requests/atmosphere/api/actions_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    let!(:appliance)  { create(:appliance, :appliance_set => appliance_set) }
+
+
+    let!(:user2) { create(:user) }
+    let!(:appliance_set2)  { create(:appliance_set, :user => user2) }
+    let!(:appliance2)  { create(:appliance, :appliance_set => appliance_set2) }
+
+    let!(:action_log) { create(:action_log, :message => message) }
+    let!(:action)  { create(:action, :appliance => appliance, :action_logs => [action_log]) }
+    let!(:other_action)  { create(:action, :appliance => appliance2) }
```
mkasztelnik: You know that `let!` is creating object in DB before every spec - even if it is not used in concrete test case. It is better to create only minimum set of DB objects which are really needed for the spec.

app/models/atmosphere/action.rb
```diff
@@ -0,0 +1,13 @@
...
+    belongs_to :appliance,
+               class_name: 'Atmosphere::Appliance'
+
+    has_many :action_logs,
+             dependent: :destroy,
+             class_name: 'Atmosphere::ActionLog'
+
+
+  end
+end
```
mkasztelnik: We were talking about something like `action.info('some msg')`, `action.warn('some warn')`, `action.error('error')` methods created in action and used to log action evens.

app/models/atmosphere/action.rb
bwilk: You mean "future" i guess?... A relation to Appliance Type definitely can be implemented i n the future just as a nullable reference to any other entity. I don't understand what you mean by PR

app/serializers/atmosphere/action_log_serializer.rb
bwilk: I suspect you meant "appliance_id" - in fact it should be there but it's not. 
There is no such field as "action_id" - I think "id" works that way.

app/controllers/atmosphere/api/v1/actions_controller.rb
bwilk: You are absolutely right...

spec/requests/atmosphere/api/actions_spec.rb
bwilk: OK

app/models/atmosphere/action.rb
bwilk: On my way :)

app/models/atmosphere/action.rb
mkasztelnik: Yeach future :D 

PR = push request

app/serializers/atmosphere/action_log_serializer.rb
mkasztelnik: Are you sure? Action has many ActionLogs, thus inside ActionLog there is `action_id` (see line 22 in schema.db)

app/models/atmosphere/action.rb
```diff
@@ -0,0 +1,24 @@
...
+  class Action < ActiveRecord::Base
+
+    belongs_to :appliance,
+               class_name: 'Atmosphere::Appliance'
+
+    has_many :action_logs,
+             dependent: :destroy,
+             class_name: 'Atmosphere::ActionLog'
+
+    def log(message, level = :info)
```
mkasztelnik: Yeap, message should be added here. What is more you can use more user friendly syntax here:

```ruby
action_logs.create(message: message, log_level: level)
```

It will automatically set correct relation between `action` and `action_log`

spec/requests/atmosphere/api/actions_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    before do
+      expect(Atmosphere::Optimizer).to receive(:instance).at_least(:once) { optimizer }
+      expect(optimizer).to receive(:run).at_least(:once).with(anything())
+    end
+
+    let!(:message) { "message123" }
+
+    let!(:user) { create(:user) }
+    let!(:appliance_set)  { create(:appliance_set, :user => user) }
```
mkasztelnik: Please use 1.9 ruby syntax when possible (`user: user`)

app/models/atmosphere/action.rb
paoolo: PR = pull request?

mkasztelnik: yeap

Nuanda:
And please remove all Hound comments after commiting the revised code :). It makes it a lot easier to see what happens here.

app/serializers/atmosphere/action_log_serializer.rb
bwilk: Ok my bad. That being said the answer to your question is different :). action_id is not needed as the action log is embedded in action serializer.

mkasztelnik:
@bwilk today you mentioned that all issues reported during code review were fixed. I don't see any changes - did you push latest version into github?

mkasztelnik:
Tested on devel - no problems. For sure we will need to add more detailed information into action log for initial optimization and scaling, but introduced mechanism works for me -> :+1:

app/models/atmosphere/action.rb
```diff
@@ -0,0 +1,22 @@
...
+  class Action < ActiveRecord::Base
+    belongs_to :appliance,
+               class_name: 'Atmosphere::Appliance'
+
+    has_many :action_logs, -> { order :id },
+             dependent: :destroy,
+             class_name: 'Atmosphere::ActionLog'
+
+    def log(message, level = :info)
+      action_logs.create(message: message, log_level: level)
```
paoolo: Where is `message` stored in `ActionLog`?

app/services/atmosphere/optimizer.rb
```diff
@@ -14,6 +14,10 @@ def run(hint)
 
     #private
     def satisfy_appliance(appliance)
+      action = Action.create(appliance: appliance,
```
paoolo: Creating new entity about `Action` is performed by creating `Action` object and performing method `.log()`?

paoolo:
Looks okey.

app/models/atmosphere/action.rb
```diff
@@ -0,0 +1,22 @@
+module Atmosphere
+  class Action < ActiveRecord::Base
+    belongs_to :appliance,
+               class_name: 'Atmosphere::Appliance'
+
+    has_many :action_logs, :order => 'id DESC'
+             dependent: :destroy,
```
paoolo: Whops?!

app/models/atmosphere/action.rb
```diff
@@ -0,0 +1,22 @@
+module Atmosphere
+  class Action < ActiveRecord::Base
+    belongs_to :appliance,
+               class_name: 'Atmosphere::Appliance'
+
+    has_many :action_logs, -> { order(id: :asc) },
+             dependent: :destroy,
```
nowakowski: Er, do we actually want to destroy the log upon destroying the appliance?

app/models/atmosphere/action.rb
```diff
@@ -0,0 +1,22 @@
...
+    end
+
+    def warn(message)
+      log(message, :warn)
+    end
+
+    def error(message)
+      log(message, :error)
+    end
+  end
```
nowakowski: Missing `info(message)`

app/models/atmosphere/action.rb
mkasztelnik: We were discussing about it with @bwilk and decided that right now yes. In the feature we can use e.g. [paranoia](https://github.com/radar/paranoia) for soft appliance and actions destroy. Appliances older than e.g. month will be removed permanently.

db/migrate/20150112083118_create_atmosphere_action_logs.rb
```diff
@@ -0,0 +1,10 @@
+class CreateAtmosphereActionLogs < ActiveRecord::Migration
+  def change
+    create_table :atmosphere_action_logs do |t|
+      t.string :message
+      t.string :log_level
+      t.references :action, index: true
+      t.timestamps
```
nowakowski: Action log doesn't need `updated_at`; suggest changing this to:
`t.datetime :created_at`

app/models/atmosphere/action.rb
nowakowski: = == ==? ;)

app/models/atmosphere/action.rb
nowakowski: I'm not actually sure I understand the logic here. There's a `dependent: :destroy` on Appliance -> Action and likewise on Action -> ActionLog so once the appliance is gone, so are all its logs. Are you *absolutely sure* that's what you want to happen? It seems really weird; I would have expected a `dependent: :nullify` here...
Regarding your other comment - I don't think automatically destroying any appliances (regardless of their age) is a good idea. In Share people have been running appliances for months and still end up needing them.

mkasztelnik: I was not clear. After appliance is destroyed by the user it can been flagged as destroyed, but still present in DB for further, eventually diagnostics. No automatic active appliance destroying :smile:

app/models/atmosphere/action.rb
paoolo: @nowakowski: Yess :smile:

app/services/atmosphere/optimizer.rb
mkasztelnik: No. `log` creates new `ActionLog` correlated with concrete `Action`.

db/migrate/20150112083118_create_atmosphere_action_logs.rb
mkasztelnik: @nowakowski good catch. What is more `ActionLog` should be immutable. You can [read here](http://stackoverflow.com/questions/3342806/turn-off-updated-at-column-in-rails) how to make the object readonly (but unlike presented example we want `ActionLog` to be remmovable).

spec/models/atmosphere/action_log_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'rails_helper'
+
+module Atmosphere
+  RSpec.describe ActionLog do
+    pending "add some examples to (or delete) #{__FILE__}"
```
mkasztelnik: You can add here specs testing `ActionLog` read only feature.

spec/models/atmosphere/action_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'rails_helper'
+
+module Atmosphere
+  RSpec.describe Action do
+    pending "add some examples to (or delete) #{__FILE__}"
```
mkasztelnik: Specs for testing `log`, `warn`, `error` methods.

app/models/atmosphere/action.rb
bwilk: log(message) does that

app/models/atmosphere/action.rb
bwilk: With @mkasztelnik we canme to the conclusion that for now it stays as it is

app/services/atmosphere/optimizer.rb
bwilk: Everything is righ @paoolo. At this place satisfy_appliance represents whole new action so a new Action is created. In the body of this method you can call log several times on that object.

db/migrate/20150112083118_create_atmosphere_action_logs.rb
bwilk: Ok. I modified migration.

bwilk:
To me everything seems fixed. Please rise your hands (1 more needed) to let mi finish this job @paoolo @nowakowski @tbartynski

mkasztelnik:
@paoolo, @nowakowski, @tbartynski can you review these changes and can we merge it and close it today?

paoolo:
For me looks good, but I guess that something is hidding if so many conversation was made in this PR which has only 0.5k lines changed and most in test. So I would like to say :+1: if this PR will only add action log support and has no impact to other things.

nowakowski:
:+1:

PR #108, opened 2015-02-10
testing ruby 2.1.4

mkasztelnik:


mkasztelnik:

nowakowski:
:+1:

nowakowski:
Works for me (already tested in my branch).

tbartynski:
I'm using rvm and with Ruby 2.1.4 both guard and rspec fails. If I set Ruby 2.1.2 in this branch I can run tests. I can't give plus one since it does not work for me but it is not a blocker from my point of view.

mkasztelnik:
Ok so merging since it helps travis and in my local env.

PR #111, opened 2015-02-10
Correct exception name

tbartynski:
Before we started using Rails engine we had an exception Air::UnsuportedOperation. Please notice the typo in exception class name. After we switched to engine the module of this exception was renamed to Atmosphere, however the code that raises this exception was not refactored accordingly. This PR fixes this bug.

### DIFF
--- a/app/exceptions/atmosphere/unsupported_exception.rb
+++ b/app/exceptions/atmosphere/unsupported_exception.rb
@@ -1,0 +1,3 @@
+module Atmosphere
+  class UnsupportedException < Exception; end
+end
--- a/app/exceptions/atmosphere/unsuported_exception.rb
+++ b/app/exceptions/atmosphere/unsuported_exception.rb
@@ -1,3 +1,0 @@
-module Atmosphere
-  class UnsuportedException < Exception; end
-end
--- a/config/initializers/unify_fog_cloud_clients.rb
+++ b/config/initializers/unify_fog_cloud_clients.rb
@@ -91,11 +91,11 @@ class Fog::Compute::AWS::Server
   end
 
   def pause
-    raise Air::UnsuportedException, 'Amazon des not support pause action'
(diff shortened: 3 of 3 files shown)


mkasztelnik:
:+1:

Nuanda:
Guys - a couple of issues:
 - this branch fails its build, it should not be set as ready for review, unless a good reason is given
 - since it fails, it should not be thumbs-upped, unless a good reason for doing so is given
 - the issue it claims to fix should not be closed, unless this branch is merged.

mkasztelnik:
@Nuanda it fails because of strange travis behavior. It was tested locally without any troubles. After switching in travis into ruby 2.1.4 everything is green.

Nuanda:
All right, I exactly meant this kind of explanation, so someone on day's leave knows what's what :). Still, the issue should not be closed until this branch is merged.

mkasztelnik:
Rebased into master to run all specs with ruby 2.1.4

mkasztelnik:
Yeap now all specs are green :heart:

app/exceptions/atmosphere/unsupported_exception.rb
```diff
@@ -0,0 +1,3 @@
+module Atmosphere
+  class UnsupportedException < Exception; end
+end
```
paoolo: What this change do? It looks like the removed file. Did you only change the name from `app/exceptions/atmosphere/unsuported_exception.rb` to `app/exceptions/atmosphere/unsupported_exception.rb` (with double `pp`)?

mkasztelnik: this is correction of very common bug in code I commit - misspell. Instead of `UnsupportedException` class was named `UnsuportedException` (one `p` missing).

paoolo:
:+1:

Issue #113, opened 2015-02-12
race condition while migrating to amazon


mkasztelnik:
It looks like amazon create image is unable to set VMT tags at creation time. This operation is no atomic. What is worse in our [amazon migration script](https://github.com/dice-cyfronet/migratio/blob/master/scripts/openstack2amazon-register.sh#L52-65) we have 30 sleep interval while checking if VMT is available on list (it is then also available for atmosphere monitoring).

Presented lack of atomicty causes bug in Atmosphere which looks as follow:

  * We have AT with VMT (`vmt1`) assigned
  * `vmt1` is migrated into Amazon as  result new `vmt2`
  * `vmt2` is discovered by Atmosphere monitoring but not yet by [migrato script](https://github.com/dice-cyfronet/migratio/blob/master/scripts/openstack2amazon-register.sh#L60-65)

As a result monitoring is not able to correlate  `vmt2` with AT (not metadata is available yet). After a while metadata is set by migrato script. Then monitoring discovers this situation and assign `vmt2` to AT - as a consequence version is incremented (`ApplianceType.version + 1`) and assigned to `vmt2`.

/cc @paoolo, @nowakowski

mkasztelnik:
Most probably fixing #112 will also fixes this issue.

mkasztelnik:
Fixed in #114

PR #114, opened 2015-02-13
setting version has precedence over setting AT

### LINKED ISSUE #112: check metadata while setting version when VMT is assigned into AT
We have following scenario:

  * VMT (`vmt1`) is available on First CS with version 1
  * it is migrated into Second CS. As a result new VMT (`vmt2`) is create (version is set to 1)
  * New AT is created and `vmt1` is assigned
  * Administrator assigns `vmt2` into AT

As a result `vmt2` version will be set into `AT.version + 1` (2). This obviously wrong. Correct solution should contact compute site, get metadata and if `source_uuid` is available than we should check if source VMT is assigned to the AT. If yes the same version should be set to new assigned VMT.

/cc @paoolo, @nowakowski

mkasztelnik:
It solved the problem when VMT is migrated into Amazon. There it is not possible to register template and set its metadata as atomic operation. As a conclusion VMT was first discovered but not assigned to appliance type and version was not set. Next after metadata was added monitoring assigns this VMT into appliance type and set version into `appliance_type.version + 1`. This is obviously bug. This commit fixes this issue, by allowing to update version by the monitoring when VMT is young. What is more during "young" period if administrator changes source VMT version it will be populated into migrated VMT.

Fixes #112 and #113

### DIFF
--- a/spec/models/atmosphere/virtual_machine_template_spec.rb
+++ b/spec/models/atmosphere/virtual_machine_template_spec.rb
@@ -314,5 +314,16 @@ describe Atmosphere::VirtualMachineTemplate do
 
       expect(vmt.version).to eq 1
     end
+
+    it 'setting version has precedence over assigning to AT' do
+      vmt = create(:virtual_machine_template, version: 15)
+      at = create(:appliance_type)
+
+      vmt.appliance_type = at
+      vmt.version = 3
+      vmt.save
+
+      expect(vmt.version).to eq 3
+    end
   end
 end
--- a/app/models/atmosphere/virtual_machine_template.rb
+++ b/app/models/atmosphere/virtual_machine_template.rb
@@ -62,7 +62,7 @@ module Atmosphere
 
     enumerize :state, in: ALLOWED_STATES
 
-    before_save :set_version, if: :appliance_type_id_changed?
+    before_save :set_version, if: :update_version?
     before_update :release_source_vm, if: :saved?
     after_update :destroy_source_vm, if: :saved?
     before_destroy :cant_destroy_non_managed_vmt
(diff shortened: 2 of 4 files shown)


mkasztelnik:
Tested on dev. @dice-cyfronet/atmo-dev-team please take a look

spec/models/atmosphere/virtual_machine_template_spec.rb
```diff
@@ -314,5 +314,16 @@ def delete_with_success_when_exception(e)
 
       expect(vmt.version).to eq 1
     end
+
+    it 'setting version has precedence between assigning to AT' do
```
nowakowski: Shouldn't it say 'over' instead of 'between'?

mkasztelnik: Good point I will fix it

mkasztelnik: Fixed

nowakowski:
:+1:

nowakowski:
(Notice how I did not wait for Travis to finish! :pray: )

mkasztelnik:
Up sorry I forgot to add information about rationale. See #112 and #113

mkasztelnik:
I will also change commit message as @nowakowski propose

mkasztelnik:
After series of @Nuanda comments (thx) I was thinking about automatic version set in callback. When setting version has precedence over setting AT `set_version` method can be simplified. Latest commit  reflects this change. 

What is more now I'm thinking about moving the logic of incrementing version into [`save`](https://github.com/dice-cyfronet/atmosphere/blob/master/app/actions/atmosphere/cloud/save.rb). I will analyze consequence of such refactoring next week and if everything will be ok new PR will be submitted. 

Now I'm waiting for this PR approval.

Nuanda:
:+1:

PR #115, opened 2015-02-17
Network interface assignment from config file

nowakowski:
Assigns a NIC to a VM create request (if defined in config file). Required by some OpenStack installations.


nowakowski:
This is needed for the PL-Grid (EGI) OST setup to work correctly (there are two NICs there and of course Fog selects the wrong one by default ;)

Nuanda:
:+1:

mkasztelnik:
Can you move this `if defined?` into [`ApplianceVmsManager`](https://github.com/dice-cyfronet/atmosphere/blob/master/app/services/atmosphere/appliance_vms_manager.rb#L93-L100) and pass it similar like other params to `VmCreator`?

nowakowski:
@mkasztelnik Done as requested.

nowakowski:
Sorry, misclicked. :)

app/services/atmosphere/appliance_vms_manager.rb
```diff
@@ -91,11 +91,20 @@ def instantiate_vm(tmpl, flavor, name)
     end
 
     def start_vm_on_cloud(tmpl, flavor, name)
+      nic = Atmosphere.nics[tmpl.compute_site.site_id]
+      if !(nic.blank?)
+        Rails.logger.info("Using custom NIC: #{nic}")
+      else
+        Rails.logger.info('Using default NIC.')
+        nic = nil
```
mkasztelnik: I believe this line can be removed

app/services/atmosphere/appliance_vms_manager.rb
```diff
@@ -91,11 +91,20 @@ def instantiate_vm(tmpl, flavor, name)
     end
 
     def start_vm_on_cloud(tmpl, flavor, name)
+      nic = Atmosphere.nics[tmpl.compute_site.site_id]
+      if !(nic.blank?)
```
mkasztelnik: `if nic` or `unless nic.blank?` - current version introduces unnecessary complexity.

lib/atmosphere.rb
```diff
@@ -97,6 +97,10 @@ def self.ability_class
     @@ability_class.constantize
   end
 
+  mattr_accessor :nics
+  mattr_reader :nics
```
mkasztelnik: This is not needed since [`mattr_accessor`](http://apidock.com/rails/Module/mattr_accessor) creates getter and setter

spec/services/atmosphere/appliance_vms_manager_spec.rb
```diff
@@ -115,7 +115,7 @@ def appliance(options)
 
     before do
       allow(vm_creator_class).to receive(:new).with(tmpl,
-        {flavor: flavor, name: name, user_data: 'user data', user_key: 'user key'})
+        {flavor: flavor, name: name, user_data: 'user data', user_key: 'user key', nic: nil})
```
mkasztelnik: In such a way this line can be simplified into:

```ruby
allow(vm_creator_class).
  to receive(:new).with(tmpl, flavor: flavor, name: name, user_data: 'user data', user_key: 'user key', nic: nil)
```

no `{}` needed

mkasztelnik:
Take a look at my comments above, when fixed :+1: from me.

mkasztelnik:
And last question: can you squash your last commits (this with typos, tests, etc.) into one commit?

mkasztelnik:
@nowakowski any progress here?

app/services/atmosphere/appliance_vms_manager.rb
nowakowski: `if nic` is not equivalent to `if !(nic.blank?)` (for example: an empty string passes `if nic` but does not satisfy `if !(nic.blank?)`.
Hound does not like `unless`.
Changed to `if nic.present?` instead.

mkasztelnik:
That's a pitty that you did not squashed "Typo", "Test", "dog", etc. into one commit, a lot o noise here :sob:

PR #118, opened 2015-02-25
Use session and CSRF protection when no API token given

mkasztelnik:
It is useful when Clew is embedded into Atmosphere home page. Than we don't need to pass any token only session cookie and valid CSRF token is needed.

CSRF token can be passed into Clew using following method `form_authenticity_token`.

To distinguish when token authentication should be used dedicated method is created in `Atmosphere::Api::ApplicationControllerExt` `token_request?`. It should return `true` when token is present in the request. It need to be overridden in every project with token support which uses Atmosphere.

### DIFF
--- a/app/controllers/atmosphere/api/application_controller.rb
+++ b/app/controllers/atmosphere/api/application_controller.rb
@@ -2,7 +2,8 @@ module Atmosphere
   module Api
     class ApplicationController < ::ApplicationController
       include Atmosphere::Api::ApplicationControllerExt
-      protect_from_forgery with: :null_session
+      protect_from_forgery with: :null_session, if: :token_request?
+      protect_from_forgery with: :exception, unless: :token_request?
 
       check_authorization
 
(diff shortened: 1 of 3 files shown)


mkasztelnik:
@dice-cyfronet/atmo-dev-team please take a look. It is needed to remove authentication token when clew is embedded in `home#index` page.

Nuanda:
:+1:

paoolo:
I am so impressed. You add 6 lines to production code and a new spec has been created.

Why `token_request?` is always return `false`.

mkasztelnik:
default `token_request?` by default return `false` because vanilla Atmosphere does not have any token authentication method defined by default. Projects which uses Atmosphere can add such authentication and then dedicated implementation of `token_request?` method need to be provided.

paoolo:
:+1:

tbartynski:
:+1:

Issue #120, opened 2015-02-25
rename atmosphere_os_families_id into os_family_id in ApplianceType


mkasztelnik:
It should be renamed to avoid this: https://github.com/dice-cyfronet/atmosphere/commit/ea3645dbd01379700db9be9392744990ed73a351

mkasztelnik:
Already fixed in #122

PR #122, opened 2015-03-04
renamed column name in appliance_types table

tbartynski:
https://github.com/dice-cyfronet/atmosphere/issues/120

It used to be atomsphere_os_famielies_id and the correct name os_family_id

### DIFF
--- a/app/models/atmosphere/appliance_type.rb
+++ b/app/models/atmosphere/appliance_type.rb
@@ -27,7 +27,7 @@ module Atmosphere
       foreign_key: 'user_id'
 
     belongs_to :os_family,
-      class_name: 'Atmosphere::OSFamily',
+      class_name: 'Atmosphere::OSFamily'
 
     has_many :appliances,
       dependent: :destroy,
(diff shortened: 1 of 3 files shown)


app/models/atmosphere/appliance_type.rb
```diff
@@ -28,7 +28,7 @@ class ApplianceType < ActiveRecord::Base
 
     belongs_to :os_family,
       class_name: 'Atmosphere::OSFamily',
-      foreign_key: 'atmosphere_os_families_id'
+      foreign_key: 'os_family_id'
```
mkasztelnik: It is necessary since it is default?

mkasztelnik:
I believe you forgot about one additional change which need to be done [here](https://github.com/dice-cyfronet/atmosphere/blob/69ae0ea7412c20662eacb7e6e3299ee78cec7c16/app/models/atmosphere/appliance_type.rb#L189)

mkasztelnik:
:+1:

bwilk:
:+1:

PR #129, opened 2015-03-24
Ignore second reboot conflict exception

### LINKED ISSUE #100: Exception is thrown when user tries to restart VM twice.
```
Excon::Errors::Conflict: Expected([200, 202]) <=> Actual(409 Conflict)
excon.error.response
  :body          => "{\"conflictingRequest\": {\"message\": \"Cannot 'reboot' while instance is in task_state reboot_started\", \"code\": 409}}"
  :headers       => {
    "Content-Length"       => "114"
    "Content-Type"         => "application/json; charset=UTF-8"
    "Date"                 => "Mon, 02 Feb 2015 11:18:03 GMT"
    "X-Compute-Request-Id" => "req-xxx"
  }
  :local_address => "xxx.xxx.xxx.xxx"
  :local_port    => 38793
  :reason_phrase => "Conflict"
  :remote_ip     => "xxx.xxx.xxx.xxx"
  :status        => 409
```

Return correct response to the user instead of 500

mkasztelnik:
We can ignore it since reboot is already in progress.

Fix #100

### DIFF
--- a/app/models/atmosphere/virtual_machine.rb
+++ b/app/models/atmosphere/virtual_machine.rb
@@ -237,8 +237,8 @@ module Atmosphere
       errors.add :base, 'Virtual Machine is not managed by atmosphere' unless managed_by_atmosphere
     end
 
-    def cloud_action(aciton_name, sucess_state)
-      action_status = cloud_server.send(aciton_name)
+    def cloud_action(action_name, sucess_state)
+      action_status = cloud_server.send(action_name)
       change_state_on_success(action_status, sucess_state)
     end
 
--- a/app/models/atmosphere/virtual_machine.rb
+++ b/app/models/atmosphere/virtual_machine.rb
@@ -98,6 +98,8 @@ module Atmosphere
 
     def reboot
       cloud_action(:reboot, :reboot)
+    rescue Excon::Errors::Conflict
+      # ok reboot already in progress
     end
 
     def stop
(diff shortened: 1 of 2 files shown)


mkasztelnik:
Tested on dev

app/models/atmosphere/virtual_machine.rb
```diff
@@ -235,23 +237,11 @@ def cant_destroy_non_managed_vm
...
     end
 
-    def monitoring_client
-      Atmosphere.monitoring_client
-    end
-
-    def cloud_action(action_name, sucess_state)
-      action_status = cloud_server.send(action_name)
+    def cloud_action(aciton_name, sucess_state)
+      action_status = cloud_server.send(aciton_name)
```
tbartynski: Rename acition_name to action_name

mkasztelnik: done

app/models/atmosphere/virtual_machine.rb
```diff
@@ -98,6 +98,8 @@ def uuid
 
     def reboot
       cloud_action(:reboot, :reboot)
+    rescue Excon::Errors::Conflict
+      # ok reboot already in progress
     end
```
tbartynski: Would it be possible to check that vm is already being rebooted based on vm model state kept in Atmo DB? I am aware that this state is updated periodically but maybe our system should not allow to reboot a vm that it thinks is being rebooted (or in booting state). What do you think?

mkasztelnik: I can be a race condition. We are changing state after the request to compute site is finished. When 2 requests will arrive in more less the same time than appliance will have `active` state in both cases.

tbartynski:
:+1:

mkasztelnik:
@Nuanda corrected in 2432915bd4eaefdbd3d8d9ba54b2a24cfa044003

Nuanda:
:+1:

PR #132, opened 2015-03-28
Get nics from ldap null client

tbartynski:


Nuanda:
Please, remove the hound comments that are not relevant and fix the others.

lib/generators/templates/atmosphere.rb
```diff
@@ -107,6 +107,11 @@
   #
   # config.metrics_store = Atmosphere::Monitoring::NullMetricsStore.new
 
+  # Defines the client that provides information about the network id
+  # for instantiated vm.
+  # Possible values:
+  #   - Atmosphere::NicProvider::NullNicProvider.new
```
mkasztelnik: I believe this line is missing:

```ruby
nic_provider_class = 'Atmosphere::NicProvider::NullNicProvider`
```

and in the line above `.new` should not be there

spec/lib/atmosphere_spec.rb
```diff
@@ -58,4 +58,39 @@ def register_cloud_cient
...
+      Atmosphere.nic_provider_class_name = 'String'
+
+      expect(Atmosphere.nic_provider_class).to eq String
+
+      # clean up
+      Atmosphere.nic_provider_class_name = nil
+    end
+
+
+    it 'returns null nic provider client when no configuration' do
```
mkasztelnik: Why this and test below are duplicated?

lib/atmosphere.rb
```diff
@@ -97,8 +97,12 @@ def self.ability_class
     @@ability_class.constantize
   end
 
-  mattr_reader :nics
-  @@nics = {}
+  def self.nic_provider(compute_site)
+    class_name = compute_site.nic_provider_class_name
+    nic_provider_class = class_name ? class_name.constantize :
+      Atmosphere::NicProvider::NullNicProvider
+    nic_provider_class.new(compute_site.nic_provider_config)
```
mkasztelnik: What if `class_name` is invalid and exception is thrown?

db/migrate/20150410124111_add_nic_provider_config_to_compute_site.rb
```diff
@@ -0,0 +1,5 @@
+class AddNicProviderConfigToComputeSite < ActiveRecord::Migration
+  def change
+    add_column :atmosphere_compute_sites, :nic_provider_config, :text
+  end
+end
```
mkasztelnik: I recommend to merge this migration with `AddNicProviderClassNameToComputeSite`. One without another IMHO does not have sense

spec/lib/atmosphere_spec.rb
```diff
@@ -58,4 +58,38 @@ def register_cloud_cient
...
+        expect(Atmosphere.nic_provider(cs).class).to eq DummyNicProvider
+      end
+
+      it 'creates provider with compute site specific configuration' do
+        PROVIDER_CONF = 'Nic provider configuration'
+        cs.nic_provider_config = PROVIDER_CONF
+        expect(DummyNicProvider).to receive(:new).with PROVIDER_CONF
+        Atmosphere.nic_provider(cs)
+      end
+    end
```
mkasztelnik: Additional spec checking what will happened when `nic_provider_class_name` is invalid would be useful.

db/migrate/20150410124111_add_nic_provider_config_to_compute_site.rb
tbartynski: We've had a discussion with @mkasztelnik  about it and decided not to change it.

lib/atmosphere.rb
mkasztelnik: I already mentioned about it in plgrid PR but just to remember: Maybe instead of passing string configuration can be converted from json structure into `key:value` hash and passed as NIC provider configuration?

spec/lib/atmosphere_spec.rb
tbartynski: I've added a validation to ComputeSite model so if the DB is not corrupted class name should be valid. Do you think it is good enough and solves the problem of invalid class name? If I'm not asked explicitly to handle invalid class name in Atmosphere code I won't do more :)

app/models/atmosphere/compute_site.rb
```diff
@@ -80,6 +80,8 @@ class ComputeSite < ActiveRecord::Base
               presence: true,
               inclusion: %w(openstack aws rackspace google_compute)
 
+    validate :nic_provider_class_defined, unless: "nic_provider_class_name.nil?"
```
mkasztelnik: Maybe simpler constraint can be used here: `if: :nic_provider_class_name`? All specs are green after this change.

spec/lib/atmosphere_spec.rb
mkasztelnik: IMHO it is good enough. Good job :clap:

app/models/atmosphere/compute_site.rb
tbartynski: I think that if: :nic_provider_class_name is always true because it really is if: :symbol 
which evaluates always to true so tests are still green. But I am not 100% sure about it.

lib/atmosphere.rb
tbartynski: Implemented in 8c64875037150be6fd6267ed48371b5935185d8f

mkasztelnik:
:+1:

Nuanda:
:+1:

PR #134, opened 2015-03-30
Possibility to create Google compute client

mkasztelnik:
TODOs:
  
  * [x] Create google compute client using for library
  * [x] Get flavors (`FlavorManager`)
  * [x] Get flavor data (`FlavorManager`)
  * [x] Get vms (`VmMonitoringWorker`)
  * [x] Get vm data (`VmUpdater`)
  * [x] Create tags for vm (`VmTagsCreator`)
  * [x] Get templates	 (`VmTemplateMonitoringWorker`)
  * [x] Destroy vm (`VirtualMachine`)
  * [x] Managing vm (`VirtualMachine`)
  * [ ] Save template	(`VirtualMachineTemplate`)
  * [x] Destroy template (`VirtualMachineTemplate`)
  * [x] Create vm (`VmCreator`)
  * [x] Import key pair (`UserKey`)
  * [x] Delete key pair (`UserKey`)
  * [x] Get template data (`VmtUpdater`)
  * [x] Permit all for http/https/tcp/udp
  * [x] Initial configuration

### DIFF
--- a/app/controllers/atmosphere/api/v1/appliance_types_controller.rb
+++ b/app/controllers/atmosphere/api/v1/appliance_types_controller.rb
@@ -27,37 +27,22 @@ module Atmosphere
         end
 
         def create
-          appl = Appliance.find appliance_type_params['appliance_id'] if appliance_type_params['appliance_id']
-          tmpl = nil
-          vm = nil
+          appl = appliance_type_params['appliance_id'] &&
+                 Appliance.find(appliance_type_params['appliance_id'])
+
           if appl
             authorize!(:save_vm_as_tmpl, appl)
-            vm = appl.virtual_machines.first
-            raise Atmosphere::Conflict.new("It is not allowed to save application twice") if vm && vm.state.saving?
+            check_for_conflict!(appl)
           else
             unless current_user.admin?
-              raise ActionController::ParameterMissing.new('appliance_id parameter is missing')
+              raise ActionController::ParameterMissing,
+                    'appliance_id parameter is missing'
             end
           end
-          begin
-            Atmosphere::ApplianceType.transaction do
-              tmpl = Atmosphere::VirtualMachineTemplate.create_from_vm(vm, appliance_type_params[:name]) if vm
 
-              new_at_params = appliance_type_params.dup
-              new_at_params['user_id'] = new_at_params.delete('author_id')
-
-              @appliance_type = Atmosphere::ApplianceType.create_from(appl, new_at_params)
-              @appliance_type.virtual_machine_templates << tmpl if tmpl
-              @appliance_type.author = current_user if @appliance_type.author.blank?
-
-              @appliance_type.save!
-            end
-          rescue
-            if tmpl and tmpl.id_at_site
-              tmpl.perform_delete_in_cloud
-            end
-            raise $!
-          end
+          @appliance_type =
+            Atmosphere::SaveAsService.new(current_user, appl,
+                                          appliance_type_params).execute
 
           render json: @appliance_type, serializer: ApplianceTypeSerializer, status: :created
         end
@@ -92,6 +77,14 @@ module Atmosphere
         end
 
         private
+
+        def check_for_conflict!(appl)
+          vm = appl.virtual_machines.first
+          if vm && vm.state.saving?
+            raise Atmosphere::Conflict,
+                  'It is not allowed to save application twice'
+          end
+        end
 
         def perform_save(appliance_id)
           appl = Atmosphere::Appliance.find(appliance_id)
--- a/app/workers/atmosphere/vm_template_monitoring_worker.rb
+++ b/app/workers/atmosphere/vm_template_monitoring_worker.rb
@@ -9,9 +9,17 @@ module Atmosphere
       begin
         logger.debug { "#{jid}: Starting VMTs monitoring for site #{site_id}" }
         site = ComputeSite.find(site_id)
-        filters = site.template_filters ? JSON.parse(site.template_filters) : nil
+        filters = site.template_filters ? JSON.parse(site.template_filters) : {}
+
         logger.debug { "#{jid}: getting images state for site #{site_id} from compute site" }
-        update_images(site, site.cloud_client.images.all(filters))
+
+        images = if filters.empty?
+                   site.cloud_client.images.all
+                 else
+                   site.cloud_client.images.all(filters)
+                 end
+
+        update_images(site, images)
         logger.debug { "#{jid}: updating VMTs finished for site #{site_id}" }
       rescue Excon::Errors::HTTPStatusError => e
         logger.error "#{jid}: Unable to perform VMTs monitoring job: #{e}"
--- a/config/initializers/unify_google_compute.rb
+++ b/config/initializers/unify_google_compute.rb
@@ -70,8 +70,8 @@ module Fog
           params[:username] = 'atmosphere'
 
           disk = create_disk(params)
-          params[:disk] = disk.
-            get_object(true, true, 'created_by_atmosphere', true)
+          params[:disk] = disk.get_object(true, true,
+                                          'created_by_atmosphere', true)
 
           server = nil
           if params[:atmo_user_key]
@@ -133,11 +133,6 @@ module Fog
 
         def flavor_id
           machine_type.split('/').last
-        end
-
-        def public_ip_address
-          # kind a hacky but good enough for POC
-          network_interfaces.first['accessConfigs'].first['natIP']
         end
 
         def created
--- a/config/initializers/unify_google_compute.rb
+++ b/config/initializers/unify_google_compute.rb
@@ -1,5 +1,8 @@
 require 'fog/google/models/compute/flavor'
 require 'fog/google/models/compute/image'
+require 'fog/google/models/compute/servers'
+require 'fog/google/models/compute/server'
+require 'fog/google/models/compute/disk'
 
 module Fog
   module Compute
@@ -19,6 +22,10 @@ module Fog
 
         def supported_architectures
           'x86_64'
+        end
+
+        def id
+          name
         end
       end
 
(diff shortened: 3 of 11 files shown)


mkasztelnik:
Example compute site configuration: 

```json
{"provider":"Google","google_key_location":"p12_path","google_key_string":"xxx","google_project":"project_id","google_client_email":"xxx@developer.gserviceaccount.com"}
```

mkasztelnik:
@nowakowski it seems like google can charge for every 10 minutes of VM usage: https://cloud.google.com/compute/pricing#machinetype - what do you think about it? Should we stick to 1h charging interval or add some extension to support 10 minutes interval?

mkasztelnik:
@nowakowski it is even better, after 10 minutes you are charged in 1 minute intervals...

nowakowski:
Hm, this is kind of bad - charge interval is a config variable and it can be changed to 10 minutes without any problems but it applies to all of Atmosphere and not to any specific compute site. :( Keep 1 hour for now.

mkasztelnik:
to create new server:

```ruby
disk_hsh = {
   :name => "disk name can be the same as server name",
   :size_gb => 10, # smallest possible value
   :zone_name => "us-central1-a",
   :source_image => "debian-7-wheezy-v20140408"
}

disk = client.disks.create(disk_hsh)
disk.wait_for { disk.ready? }
disks = [disk]

server_hsh = {
 :name=>"mktest",
 :machine_type=>"n1-standard-1",
 :zone_name=>"us-central1-a",
 :username=>"marek",
 :public_key_path=>"/home/marek/.ssh/id_rsa.pub",
 :disks=> disks
}

server = client.create(server_hsh)
```

There is also possible to set that disk should be deleted while server is terminated - need to check how to set it.

mkasztelnik:
Creating image from instance:

* Instance **need** to be deleted before creating new image. Disk **cannot** be attached to any instance

```ruby
disk = ... # disk fog object
client = ... # fog google compute client

disk_uri = "#{disk.zone}/disks/#{disk.name}"
img = client.insert_image('name', description: 'description', sourceDisk: disk_uri)
```

* At the end we can start new instance, but it will receive different ip address.

nowakowski:
@mkasztelnik Could you be a bit more specific? We're not intimately familir with the Google Compute cloud management model. In particular, I would like to ask the following questions:
  - Does your first bullet list the conditions which need to be satisfied in order for the image creation request to succeed?
  - If the above is true, how should one interpret the notion that the "instance needs to be deleted"? Do you mean suspended?
  - If the instance must be physically destroyed to feed the image creation process, what happens if the image cannot be successfully created? Can the instance be recovered?
  - Does the GC virtualization model entail storage of all instance data (OS + userspace) on a single disk image which is preserved in the cloud, or is the instance assembled from a separate OS image and userspace data like e.g. in Amazon EC2?
  - What is a "disk zone"?

In general, if the creation of an image requires destruction of an instance (which seems relly weird), I would suggest asking DH to provide an appropriate warning popup ("molly guard") whenever someone attempts to save a ApplianceType based on an existing devmode instance which is backended by a VM which runs on a cloud site of type 'google' - and the proceeding with the instance deletion/image creation request regardless.

mkasztelnik:
Simply google compute has similar solution as Azure. First you are creating disk with OS, next you "attache" CPU and RAM into the disk. Google compute understand instance deleting by removing CPU and RAM from the machine. Persistent storage with OS data are stored on disk, which you need to manage separately.

So to save Appliance as Appliance Set in our case we need to delete Virtual Machine without destroying disk attached to this Virtual Machine. Only then we can create image (aka. atmosphere Virtual Machine Template). The problem is connected with the way how atmosphere works. It assumes that VM attached into appliance always exists (even when it is in saving mode). This is not true in Google compute case.

mkasztelnik:
@nowakowski, @tbartynski, @Nuanda, @bwilk please read last 3 comments and if you have proposition how it can be solved in atmosphere (not in GUI - we will deal with it later on) please leave a comment :smile:

mkasztelnik:
I propose to make this PR as ready for review, make the review process merge it and then create 2 additional PRs: first one with "save as" capability and the second one with "save" capability.

@dice-cyfronet/atmo-dev-team please take a look at it.

mkasztelnik:
@dice-cyfronet/atmo-dev-team  unfortunately we need to make it very fast, because my free google compute grant finishes after 7 days - so please treat this PR as priority task on your TODO list.

config/initializers/unify_google_compute.rb
```diff
@@ -0,0 +1,207 @@
...
+        end
+
+        private
+
+        def create_disk(params)
+          disk_params = {
+            name: params[:name],
+            # think about taking it from flavor or user requirement.
+            # Flavor disk is currently set to 0 so this also need to be
+            # rethinked.
```
nowakowski: rethought ;)

app/workers/atmosphere/vm_template_monitoring_worker.rb
```diff
@@ -9,9 +9,17 @@ def perform(site_id)
...
+        filters = site.template_filters ? JSON.parse(site.template_filters) : {}
+
         logger.debug { "#{jid}: getting images state for site #{site_id} from compute site" }
-        update_images(site, site.cloud_client.images.all(filters))
+
+        images = if filters.empty?
+                   site.cloud_client.images.all
+                 else
+                   site.cloud_client.images.all(filters)
+                 end
```
Nuanda: Why can't we call images(filters) when filters == {} anymore?

mkasztelnik: Because Google compute `all` method does not allow any parameter.

config/initializers/unify_google_compute.rb
```diff
@@ -0,0 +1,207 @@
...
+
+        alias_method :status_orig, :status
+        def status
+          s = status_orig
+
+          if s == 'READY'
+            'active'
+          else
+            s
+          end
```
Nuanda: maybe: `status_orig == 'READY' ? 'active' : status_orig`

Nuanda:
:+1:

app/controllers/atmosphere/api/v1/appliance_types_controller.rb
```diff
@@ -93,6 +78,14 @@ def endpoint_payload
 
         private
 
+        def check_for_conflict!(appl)
+          vm = appl.virtual_machines.first
+          if vm && vm.state.saving?
+            raise Atmosphere::Conflict,
+                  'It is not allowed to save application twice'
```
nowakowski: Może "This service is already being saved. Please wait for this operation to conclude before attempting to save the service again."

config/initializers/unify_google_compute.rb
```diff
@@ -0,0 +1,207 @@
...
+        def create_disk(params)
+          disk_params = {
+            name: params[:name],
+            # think about taking it from flavor or user requirement.
+            # Flavor disk is currently set to 0 so this also need to be
+            # rethinked.
+            size_gb: 10,
+            zone_name: params[:zone],
+            source_image: params[:image_id],
+            autoDelete: true
```
nowakowski: Is camel case appropriate here?

config/initializers/unify_google_compute.rb
```diff
@@ -0,0 +1,207 @@
...
+        end
+
+        def stop
+          raise Atmosphere::UnsupportedException,
+                'Fog des not support google compute stop action'
+        end
+
+        def pause
+          raise Atmosphere::UnsupportedException,
+                'Google compute des not support pause action'
```
nowakowski: ...does...

config/initializers/unify_google_compute.rb
```diff
@@ -0,0 +1,207 @@
...
+          Time.parse(creation_timestamp)
+        end
+
+        def id
+          name
+        end
+
+        def stop
+          raise Atmosphere::UnsupportedException,
+                'Fog des not support google compute stop action'
```
nowakowski: ...does...

config/initializers/unify_google_compute.rb
```diff
@@ -0,0 +1,207 @@
...
+        end
+
+        def pause
+          raise Atmosphere::UnsupportedException,
+                'Google compute des not support pause action'
+        end
+
+        def suspend
+          raise Atmosphere::UnsupportedException,
+                'Google compute des not support suspend action'
```
nowakowski: ...does...

config/initializers/unify_google_compute.rb
```diff
@@ -0,0 +1,207 @@
...
+        end
+
+        def start
+          raise Atmosphere::UnsupportedException,
+                'Fog des not support google compute start action'
```
nowakowski: ...does...

app/controllers/atmosphere/api/v1/appliance_types_controller.rb
mkasztelnik: This is original message, I did not change here anything. If you believe this message should be improved please do it in another PR.

config/initializers/unify_google_compute.rb
mkasztelnik: I need to be camel case because it is converted into json body request send directly into google compute and google compute requires camel case.

config/initializers/unify_google_compute.rb
mkasztelnik: Will be fixed in my next commit :smile:

app/controllers/atmosphere/api/v1/appliance_types_controller.rb
nowakowski: @mkasztelnik Hmm, I've often received PR comments regarding sections of code which were not actually part of the pending PR. Many of those were from you - and I've never made a fuss about it. Let's not overbureaucratize, shall we? ;)

mkasztelnik: Fixed in https://github.com/dice-cyfronet/atmosphere/pull/138

nowakowski:
:+1:

PR #140, opened 2015-04-08
Save at google compute

mkasztelnik:
Snapshot is used to save VM as VMT. Snapshots on google compute can be created without destroying instances. Additionally snapshots are a little bit cheaper than images.

### DIFF
--- a/config/initializers/unify_google_compute.rb
+++ b/config/initializers/unify_google_compute.rb
@@ -192,8 +192,7 @@ module Fog
 
       class Server
         def image_id
-          disk_id = bootable_disk['source'].split('/').last
-          service.disks.get(disk_id).image_id
+          service.disks.get(bootable_disk_name).image_id
         end
 
         def task_state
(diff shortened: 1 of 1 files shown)


mkasztelnik:

mkasztelnik:
This is an implementation of last missing bullet from #134

tbartynski:
@mkasztelnik Have you checked what happens when you instantiate a vm from template A, take a snapshot of vm, destroy vm and destroy template A. Is the snapshot still usable?

mkasztelnik:
Yeap. Snapshot is fully functional image, without reference to its parent but a little bit cheaper and private (visible only for specified project).

config/initializers/unify_google_compute.rb
```diff
@@ -100,15 +150,31 @@ def create_disk(params)
...
         end
 
+        def snapshot?(id_at_site)
+          id_at_site.start_with?('snapshot:')
+        end
+
+        def real_id(id_at_site)
+          match = id_at_site.match(/\A.*:(.*)\z/)
+          match ? match[1] : id_at_site
+        end
```
tbartynski: There are 2  identical definitions of snapshot? and real_id methods. Could you define it in one place and reuse it?

mkasztelnik: Good catch. Improved in a289ebcd20985071f7f72660d4ed5a1831f7e336

tbartynski:
:+1:

config/initializers/unify_google_compute.rb
```diff
@@ -100,9 +154,16 @@ def create_disk(params)
             # rethought.
             size_gb: 10,
             zone_name: params[:zone],
-            source_image: params[:image_id],
             autoDelete: true
```
Nuanda: Again, is this mixture of snake and camel case required by the external API?

mkasztelnik: I already answered similar question here https://github.com/dice-cyfronet/atmosphere/pull/134#discussion_r27894341 :smile:. It is required because some of the parameter is used by fog client and other are directly translated into json structure sent to google compute.

Nuanda:
:+1:

PR #139, opened 2015-04-07
Up to ruby 2.2.1

mkasztelnik:
This is a first upgrade step. Next steps are as follow:
  * upgrade into latest rails 4.1.x and latest compatible gems 
  * upgrade into rails 4.2

PR #141, opened 2015-04-09
Ruby 4.1.10 and other gems upgrade

mkasztelnik:
This is an second step of atmosphere upgrade. First one is presented in #139 (this PR also contains changes from #139). More noticing changes are connected with rails upgrade into latest version from 4.1 line and ffaker upgrade into 2.0.0 (namespace is changed from `faker` into `ffaker`).

/cc @nowakowski, @tbartynski, @Nuanda please take a look, test is with the project you are responsible for. I will be preparing vphshare version and deploying it into vph-dev env.

TODOs:
  * [x] @mkasztelnik check if it works with vph
  * [x] @tbartynski, @Nuanda, @nowakowski check if it works with plgrid
  * [x] @mkasztelnik check if it works with ismop


mkasztelnik:
This branch is deployed into `vph-dev` and `ismop` envs for testing. So far so good - no issues discovered.

mkasztelnik:
This update is also important because of http://danlec.com/blog/bug-in-sundown-and-redcarpet

Nuanda:
Do we use markup parsing of user-provided data anywhere?

mkasztelnik:
No, that is why it is not so critical, but worth to update :smile:.

nowakowski:
Appears to work for plgrid but I'd still like @tbartynski to confirm.

tbartynski:
Dependencies should also be updated in atmosphere.gemspec.

mkasztelnik:
@tbartynski done in d070121

tbartynski:
I did some tests. I created deps_upgrade branch in plgrid https://gitlab.dev.cyfronet.pl/atmosphere/plgrid/tree/deps_upgrade
See changes there.
I am not able to test cloud related features due to firewalls but I started the web app and logged in using OpenId. I would give it a go but we must keep in mind that https://gitlab.dev.cyfronet.pl/atmosphere/plgrid/merge_requests/27 needs to be merged also.

:+1:

mkasztelnik:
@Nuanda, @nowakowski can you take a look at this PR?

Nuanda:
Why do we switch from Faker to FFaker?

mkasztelnik:
In ffaker `2.0.0` they change namespace from Faker into FFaker to allow usage of both ffaker and faker in one project. Take a look [here](https://github.com/EmmanuelOga/ffaker/blob/master/Changelog.md) for details.

Nuanda:
Right, I see. They made good decision :). Recently in another project I went back to Faker since it has a richer library of value generators.

But here, doesn't that change make Atmo ffaker 2.0.0 dependent? I mean, shouldn't it be required in gemspec with exact version number?

mkasztelnik:
@Nuanda I choose ffaker at the beginning of Atmo development because it was much more faster than faker. Recently I wrote on some blog post that speed of faker and ffaker is now comparable. I was thinking about the switch but did not want to make to much changes. 

You are right about Atmo ffaker 2.2.0 dependent. This is tricky issue. I created separate issue to track this problem - #153.

Nuanda:
:+1:

PR #146, opened 2015-04-13
Services

mkasztelnik:
In the scope of this PR I'm planning to unify how services looks like in atmosphere. Basically following conditions should be met:

  * Service is doing only **one** thing and has only **one** public method `execute`
  * All needed parameters are passed using constructor parameters

So as a conclusion every atmosphere service will have following structure:

```ruby
class MyService
  def initialize(set, of, parameters)
    ...
  end

  def execute
    # do actual job
  end
end
```

List of class we currently have in `app/services' directory:
  * [x] cloud/vm_creator
  * [x] cloud/vm_tags_manager
  * [x] cloud/vmt_updater
  * [ ] monitoring/influxdb_metrics_store
  * [ ] monitoring/null_client
  * [ ] monitoring/null_metrics_store
  * [ ] monitoring/zabix_client
  * [ ] proxy/appliance_proxy_updater
  * [ ] proxy/compute_site_appliances_updater
  * [ ] proxy/compute_site_url_updater
  * [ ] proxy/url_generator
  * [x] act_copier
  * [ ] affected_appliance_aware_manager
  * [ ] billing_service
  * [ ] dnat_wrangler
  * [ ] flavor_manager
  * [ ] optimizer
  * [x] pmt_copier
  * [x] save_as_service
  * [ ] vm_destroyer
  * [x] vm_updater
  * [x] vmt_migrator

### DIFF
--- a/app/models/atmosphere/appliance_type.rb
+++ b/app/models/atmosphere/appliance_type.rb
@@ -148,7 +148,7 @@ module Atmosphere
 
     def self.create_from(appliance, overwrite = {})
       at = ApplianceType.new appliance_type_attributes(appliance, overwrite)
-      PmtCopier.copy(appliance.dev_mode_property_set).each do |pmt|
+      PmtCopier.new(appliance.dev_mode_property_set).execute.each do |pmt|
         pmt.appliance_type = at
         at.port_mapping_templates << pmt
       end if appliance and appliance.dev_mode_property_set
--- a/app/models/atmosphere/appliance_type.rb
+++ b/app/models/atmosphere/appliance_type.rb
@@ -152,7 +152,7 @@ module Atmosphere
         pmt.appliance_type = at
         at.port_mapping_templates << pmt
       end if appliance and appliance.dev_mode_property_set
-      ActCopier.copy(appliance.appliance_type).each do |act|
+      ActCopier.new(appliance.appliance_type).execute.each do |act|
         act.appliance_type = at
         at.appliance_configuration_templates << act
       end if appliance
--- a/app/models/atmosphere/dev_mode_property_set.rb
+++ b/app/models/atmosphere/dev_mode_property_set.rb
@@ -73,7 +73,7 @@ module Atmosphere
 
       dev_mode_property_set = DevModePropertySet.new(attrs)
       dev_mode_property_set.port_mapping_templates =
-        PmtCopier.copy(appliance_type).each  do |pmt|
+        PmtCopier.new(appliance_type).execute.each  do |pmt|
           pmt.dev_mode_property_set = dev_mode_property_set
         end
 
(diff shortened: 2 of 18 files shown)


mkasztelnik:
@dice-cyfronet/atmo-dev-team I took the simplest cases just to show you how I see services. I did not change the service name, to keep amoud of changes minimal. For sure names should be changed, e.g. `UpdateVm.new(params).execute` is better than `VmUpdater.new(params).execute` - I will do this after receiving feedback from you in another PR. 

If you OK with the changes I will slowly transforming existing elements into such services, starting from Optimizer - here I'm planning to remove callbacks from `Appliance` and create `CreateAppliance`, `DestroyAppliance` services, which will create/destroy appliance and trigger optimization action. If should solve a lot of problems we currently have with testing our system.

tbartynski:
@mkasztelnik I remember we've had such a discussion in the past. In my opinion service's life-cycle is longer and it should be able to handle many request. What you are describing should be named a task or a job. It is a well-defined portion of work to be done. We should start this refactoring with creating app/tasks or app/jobs and move some of services there.

mkasztelnik:
I remember this conversation. In the meantime I browsed a lot of existing projects which uses services paradigm, read couple of blogs and it seems like rails community treat `services` folder to store something that we are calling `Actions` in Java. That is why I propose to stick to this convention. 

If Service is doing only one thing then it is simple to see what is going on in the application. E.g. i you see `SaveAs` service which takes `appliance` as a parameter than you already know what it does - this is what I really like in this approach :smile:

tbartynski:
I don't argue with advantages of such convention. The name service is counter-intuitive for me however. @Nuanda, @nowakowski what do you think?

nowakowski:
I will defer to @mkasztelnik - if he wants to call it a service then let's call it a service.

mkasztelnik:
You know service, action, etc. name is a detail... The most important element here is encapsulation of concrete business logic in one class, which does only **one** thing. If you OK with it than I will be very happy :smiley:

nowakowski:
I'm OK with that.

mkasztelnik:
@dice-cyfronet/atmo-dev-team I'm waiting for the review :)

Nuanda:
I don't think this is a matter of code review - it is rather a matter for discussing future architecture of atmosphere. I personally am reluctant to have such important changes just prior to the project meeting and, what is connected with that, PLGrid deployment testing and demo.

So, let's not merge that to master just yet, unless there is some important reason to do that.

mkasztelnik:
I took the simplest case here - so no revolution but mostly method name refactoring here. The most important thing for me is to start working on refactoring optimization process for ISMOP deployment and I agree that it will be a big thing. That is why it would be nice to have this architecture discussion sooner rather than later, can we have e.g. 1h long discussion about this issue during this or next week?

tbartynski:
@mkasztelnik if you start your work after the GA and try to make changes transparent for projects using Atmosphere engine then I give a :+1:

mkasztelnik:
@dice-cyfronet/atmo-dev-team do you think I can merge it already?

mkasztelnik:
@dice-cyfronet/atmo-dev-team if no rejection from your side I will merge it today at 12 AM

PR #151, opened 2015-04-14
Use Atmosphere engine migrations in project using Atmosphere

mkasztelnik:
See http://pivotallabs.com/leave-your-migrations-in-your-rails-engines for details.

For vph, plgrid and exp dedicated sql script which removes copied atmo migrations numbers from schema table and add origin numbers need to be created and performed before pushing this change.

This change requires dependent projects to remove old migration timestamps from DB and add original atmosphere migration timestamps. 

Below is the status of dependent projects preparation for this change:
  * [x] https://github.com/VPH-Share/atmosphere-vph/pull/10 vph (responsible @mkasztelnik)
  * [x] https://gitlab.dev.cyfronet.pl/atmosphere/plgrid/merge_requests/26 plgrid (responsible @tbartynski, @nowakowski)
  * [x] https://github.com/ismop/exp/pull/1 exp (responsible @mkasztelnik)

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,8 +99,8 @@ To upgrade to latest master run following commands:
 # update atmosphere gem
 bundle update atmosphere
 
-# install new migrations
-rake atmosphere:install:migrations
+# apply new migrations
+rake db:migrate
 ```
 
 ## Contributing
--- a/lib/atmosphere/engine.rb
+++ b/lib/atmosphere/engine.rb
@@ -39,6 +39,14 @@ module Atmosphere
       g.helper false
     end
 
+    initializer :append_migrations do |app|
+      unless app.root.to_s.match root.to_s
+        config.paths['db/migrate'].expanded.each do |expanded_path|
+          app.config.paths['db/migrate'] << expanded_path
+        end
+      end
+    end
+
     if Rails.env.test?
       initializer 'model_core.factories',
                   after: 'factory_girl.set_factory_paths' do
--- a/lib/generators/atmosphere/install_generator.rb
+++ b/lib/generators/atmosphere/install_generator.rb
@@ -20,10 +20,6 @@ module Atmosphere
         route 'mount Atmosphere::Engine => "/"'
       end
 
-      def copy_migrations
-        rake 'atmosphere:install:migrations'
-      end
-
       def show_readme
         readme 'README' if behavior == :invoke
       end


nowakowski:
@mkasztelnik Maybe it's just me but while I understand the basic concept I don't really know what rake task I'm supposed to run, when and where. Linking a devblog is not a substitute for documentation. :)

So:
- I have a vphshare checkout, a plgrid checkout and an atmosphere checkout. I can write my own SQL scripts to purge the imported migration numbers from my DBs but once I do that, how do I install any new migrations? Do I need to run `db:migrate` in my wrapper project dir or in the Atmosphere dir?
- How do I manage interleaved migrations? I assume I will need to configure my wrapper projects to look for migrations in Atmosphere's `db/migrate` dir, but how do I accomplish that?

mkasztelnik:
I described update procedure in this PR for vph deployment https://github.com/VPH-Share/atmosphere-vph/pull/10.  The most crucial thing is to update `schema_migration` versions before executing `rake db:migrate`.

To prepare rake script for other project it is enough to invoke this in your bash:

```bash
for f in `ls /path/to/other/deployment/db/migrate | grep atmosphere`
do
echo `echo "$f" | cut -d'_' -f1`
done
```
and update [this](https://github.com/VPH-Share/atmosphere-vph/blob/use_engine_migrations/lib/tasks/use_original_atmo_migrations.rake#L4-L18) part of rake task with printed timestamps.

The beauty of this solution is that you don't need to install any new migrations :smile:, rails migration path will be extended and automatically all atmosphere migrations will be added. 

According to interleaved migrations: everything is fine here since atmosphere migrations have earlier timestamps than e.g. vph project migrations. And you don't need to configure anything. It is done automatically thanks to  [this initializer](https://github.com/dice-cyfronet/atmosphere/blob/e4210212ebee7adebbae2b1a1153475d81829c45/lib/atmosphere/engine.rb#L42-L47). It extends path where project which mount atmosphere is searching migration files. As a conclusion migrations from `project/db/migrate` and `atmosphere/db/migrate` are used.

mkasztelnik:
Aaa one more thing: mentioned rake task is valid until new migration is added into atmosphere `master`. In such a way rake task need to be updated.

mkasztelnik:
@tbartynski, @nowakowski I created 3 separated branches for vph, exp and plgrid - we should coordinate deployment in these 3 projects. Is 20.04.2015 (next Monday) OK with you?

nowakowski:
@mkasztelnik Affirmative.

tbartynski:
I think this PR should not have a ready for review label since we don't really review the code. It should have a ready-for-test label.

nowakowski:
:+1:

mkasztelnik:
This PR was already deployed into `ismop` and `vph-dev` environments.

PR #156, opened 2015-04-20
reallocate_default callback condition fixed

### LINKED ISSUE #155: Random failing test  for UserFund
From time to time we have this test failed:

```
 1) Atmosphere::UserFund#reallocate_default does nothing when removing non-default fund
     Failure/Error: expect(user_fund.reload.default).to be_truthy
       expected: truthy value
            got: false
     # ./spec/models/atmosphere/user_fund_spec.rb:100:in `block (3 levels) in <top (required)>'
```

@Nuanda can you take a look at this issue?

Nuanda:
Fixes #155 

The reallocate_default callback was called improperly often. I fixed that with Proc condition. While I'm not able to prove this will fix the issue (it's an issue that is quite difficult to recreate), I believe this was the cause.

### DIFF
--- a/app/models/atmosphere/user_fund.rb
+++ b/app/models/atmosphere/user_fund.rb
@@ -24,7 +24,10 @@ module Atmosphere
               }
 
     around_save :ensure_single_default
-    after_destroy :reallocate_default if :default
+    after_destroy :reallocate_default,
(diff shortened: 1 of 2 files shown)


mkasztelnik:
:+1:

tbartynski:
:+1:

PR #158, opened 2015-04-21
Optimizer considers only funded cs

### LINKED ISSUE #157: Optimizer Default strategy may choose compute site that is not funded by user's funds
The `Manual` strategy seems to be reusing the `Default` strategy for that purpose as well (i.e. for choosing VM candidates), so fixing the `Default` is probably enough.

Nuanda:
Fixes #157


Nuanda:
@tbartynski @mkasztelnik Can you look at that fix?

mkasztelnik:
:+1:

tbartynski:
:+1:

PR #160, opened 2015-04-23
Extract destroy unused VMs into service

mkasztelnik:
The goal is to extract all logic from `Atmosphere::Optimizer`. This is first step.

PR #162, opened 2015-04-24
Extract satisfy appliance from optimizer

mkasztelnik:
This is a second step after #160 to remove optimizer from the appliance callbacks. It extracts all code specific to satisfying appliance from optimizer and move it into separate service. What is more all specs connected with satisfying appliance were extracted from `optimizer_spec` into separate service spec file. 

I'm aware that hound will have a loooooooooot of warnings here. I'm planning to skip all of them since I don't want to change anything in existing specs - I want to be sure that this refactoring will not break anything. 

Mentioned specs will be refactored when `SatisfyAppliance` refactoring will be made.

@dice-cyfronet/atmo-dev-team please take a look.


Nuanda:
This is not ready_for_review :). In order to be, all Hound comments must be either:
 - dealt with by adjusting code and commiting change, or
 - removed, if the author disagrees with a particular comment.

Such approach helps reviewers a lot.

mkasztelnik:
@Nuanda I removed all hound errors according to the comment I made in this PR description :) I will refactor the code and specs after optimization code will be removed from the appliance callbacks.

app/services/atmosphere/cloud/satisfy_appliance.rb
```diff
@@ -0,0 +1,88 @@
...
+                                  tmpl: tmpl.id_at_site,
+                                  flavor: flavor.id_at_site))
+                appl_manager.spawn_vm!(tmpl, flavor, vm_name)
+                action.log(I18n.t('satisfy_appliance.vm_started',
+                                  tmpl: tmpl.id_at_site,
+                                  flavor: flavor.id_at_site))
+              end
+            end
+          end
+          unless appl_manager.save
```
Nuanda: What happens to the VM started in line 37? Is this stopped somehow? In case appl_managed does not save properly, also the Billing action is not called. How does it influence the situation? I mean, does that cause the user to get a "free vm" running?

mkasztelnik: That is a good question :smile: I believe that user will have free VM until next billing execution is triggered. After that (s)he will be charged for previous hour. 

Of course this is a bug (https://github.com/dice-cyfronet/atmosphere/issues/174) and we should fix it. I don't want to fix is in the scope of this PR, because here I'm just moving code around files.

mkasztelnik:
@dice-cyfronet/atmo-dev-team Since it is only moving code around files if no rejection from your side I will merge it today at 12 AM

Nuanda:
:+1:

PR #170, opened 2015-05-05
vm.destroy callback stops operation when called on unmanaged vm

### LINKED ISSUE #169: It is possible to delete VMs that are not managed by Atmo
The `cant_destroy_non_managed_vm` VM model callback is setting the error message but it is not preventing the actual destroy operation from taking place.

In case this is not the intended behavior, I'll provide a fix in a PR.

Nuanda:
Fixes #169


mkasztelnik:
:+1:

tbartynski:
:+1:

PR #177, opened 2015-05-12
VM in saving state when task state is image_pending_upload

mkasztelnik:
This is a fix for: https://redmine.dev.cyfronet.pl/issues/3329. It solves following problem:

  * VM is in state active
  * Save button is clicked as a result new template should be created and state is changed into saving
  * We are canceling appliance saving process after e.g. 10s - open stack correctly removes template connected with deleted AT
  * As a result state of the appliance changes from "saving" into "active" right now. In the nova VM state is set to "active" but "task_state" is set to "image_pending_upload" - openstack does not cancel this operation for some reason. 
  * Now if we will try to save appliance once again Atmosphere throws error.

### DIFF
--- a/app/services/atmosphere/vm_updater.rb
+++ b/app/services/atmosphere/vm_updater.rb
@@ -75,8 +75,16 @@ module Atmosphere
     end
 
     def map_saving_state(vm, key)
-      (:saving if vm.saved_templates.count > 0) ||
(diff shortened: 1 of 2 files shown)


mkasztelnik:

Nuanda:
:+1:

tbartynski:
:+1:

PR #183, opened 2015-05-12
VMF model equipped with two new callbacks

### LINKED ISSUE #182: When flavor is destroyed its pricing stays behind
They should be removed. Reported by @nowakowski

Nuanda:
Fixes #182 

Also prevents VMF.destroy if there are VMs attached to that flavor.


tbartynski:
:+1:

mkasztelnik:
:+1:

PR #184, opened 2015-05-13
Load admin abilities only when user asks for it

### LINKED ISSUE #176: Owner-only ApplianceTypes are being displayed for all users
This shouldn't be.

mkasztelnik:
Fixes #176 

### DIFF
--- a/spec/requests/atmosphere/api/clew_spec.rb
+++ b/spec/requests/atmosphere/api/clew_spec.rb
@@ -13,7 +13,6 @@ describe Atmosphere::Api::V1::ClewController do
     end
 
     context 'when authenticated' do
-
       let(:user)           { create(:developer) }
       let(:different_user) { create(:user) }
       let(:admin)          { create(:admin) }
@@ -112,7 +111,6 @@ describe Atmosphere::Api::V1::ClewController do
 
         expect(json_response).to eq expected_response
       end
-
     end
 
     def clew_ai_response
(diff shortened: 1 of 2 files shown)


tbartynski:
:+1:

spec/requests/atmosphere/api/clew_spec.rb
```diff
@@ -172,20 +167,29 @@ def cs
         expect(clew_at_response['appliance_types'][1]['matched_flavor']).to clew_flavor_eq  flavor
 
         #TODO complete this test to check all the returned fields
-
       end
+    end
+
+    it 'does not return AT with visible_to owner when user is not owner' do
```
nowakowski: How does this rspec relate to the subject of the PR?

mkasztelnik: Not owned ATs were loaded because admin role was always turned on (off course for admins) for all crew requests. This was main issue for #176. This spec verifies that when not asked admin roles is not loaded in cancan abilities and not owned ATs are not returned.

nowakowski: I wonder, though - should 'owner-only' ATs really be hidden from admins? Doesn't sound very "adminlike" if you know what I mean... :)

mkasztelnik: This is a general solution for all `index` like rest services. If you want to use admin abilities you need to add `all=true` param. Take a look at https://github.com/dice-cyfronet/atmosphere/blob/master/app/controllers/atmosphere/api/application_controller.rb#L84-L86 for details.

mkasztelnik:
@nowakowski any additional comments/remarks?

nowakowski:
Nope, I'm good.

nowakowski:
:+1:

PR #185, opened 2015-05-15
Use optimizer services

mkasztelnik:
Code for scaling and destroying unused VM was removed from optimizer and appliance callbacks and moved into dedicated services.

### DIFF
--- a/app/services/atmosphere/destroy_appliance_set.rb
+++ b/app/services/atmosphere/destroy_appliance_set.rb
@@ -1,0 +1,19 @@
+module Atmosphere
+  class DestroyApplianceSet
+    def initialize(appliance_set)
+      @appliance_set = appliance_set
+    end
+
+    def execute
+      @appliance_set.appliances.each do |appliance|
+        unless DestroyAppliance.new(appliance).execute
+          @appliance_set.errors.
+            add(I18n.t('appliance_seta.cannot_remote_appliance',
+                       appliance: appliance))
+        end
+      end
+
+      @appliance_set.destroy
+    end
+  end
+end
--- a/app/services/atmosphere/destroy_appliance_set.rb
+++ b/app/services/atmosphere/destroy_appliance_set.rb
@@ -8,7 +8,7 @@ module Atmosphere
       @appliance_set.appliances.each do |appliance|
         unless DestroyAppliance.new(appliance).execute
           @appliance_set.errors.
-            add(I18n.t('appliance_seta.cannot_remote_appliance',
+            add(I18n.t('appliance_sets.cannot_remove_appliance',
                        appliance: appliance))
         end
       end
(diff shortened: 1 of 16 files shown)


mkasztelnik:

app/services/atmosphere/destroy_appliance_set.rb
```diff
@@ -0,0 +1,19 @@
...
+  class DestroyApplianceSet
+    def initialize(appliance_set)
+      @appliance_set = appliance_set
+    end
+
+    def execute
+      @appliance_set.appliances.each do |appliance|
+        unless DestroyAppliance.new(appliance).execute
+          @appliance_set.errors.
+            add(I18n.t('appliance_seta.cannot_remote_appliance',
```
nowakowski: Two typos ('seta' vs 'sets'; 'remote' vs. 'remove' - the second one requires a corresponding change in `locales/en.yml').

mkasztelnik: Fixed in https://github.com/dice-cyfronet/atmosphere/commit/0a28f732fbb2d30303c16bc7c4f3c390743691d4

app/services/atmosphere/destroy_appliance_set.rb
```diff
@@ -0,0 +1,19 @@
+module Atmosphere
+  class DestroyApplianceSet
+    def initialize(appliance_set)
+      @appliance_set = appliance_set
+    end
+
+    def execute
```
Nuanda: Can we use DB transaction on this entire execute operation? Otherwise we risk removing just a subset of Appliances from a given Set.

app/services/atmosphere/optimizer.rb
```diff
@@ -8,8 +8,6 @@ class Optimizer
     # :appliance_id DB id of the appliance that was created. Optimization only akes care of finding a vm or creating a new one for given appliance.
     def run(hint)
       satisfy_appliance(hint[:created_appliance]) if hint[:created_appliance]
-      terminate_unused_vms if hint[:destroyed_appliance]
-      scale(hint[:scaling]) if hint[:scaling]
```
Nuanda: Why have you removed these two calls?

app/services/atmosphere/destroy_appliance_set.rb
mkasztelnik: Unfortunately not. While deleting appliance destructive operation are performed on compute site (VMs are removed). If transaction would be used here and rollback would be performed we would have missing VMs. 

This is a reason I implemented it using best effort way.

app/services/atmosphere/optimizer.rb
mkasztelnik: These 2 hints are not used anymore, instead services are used - so just a cleaning.

Nuanda:
:+1:

tbartynski:
:+1:

PR #187, opened 2015-05-18
Mention about using factories in dependent projects

### LINKED ISSUE #153: Atmosphere factories requires ffaker
As @Nuanda correctly noted in https://github.com/dice-cyfronet/atmosphere/pull/141#issuecomment-93528661 Atmosphere factories [exposed to other RoR apps in test env](https://github.com/dice-cyfronet/atmosphere/pull/97) requires ffaker (after #141 will be merged ffaker 2.0.0). 

I'm not sure how such dependency can be defined in gemfile. We cannot add it as runtime dependency, because it is required only in test env. On the other hand defined it as development dependency will not solve the issue, since it is not propagated. 

The only solution I see right now is to add note that ffaker is required, if you want to use factories exposed by Atmoshere, into README. 

@dice-cyfronet/atmo-dev-team do you have any other proposition how to solve this issue?

mkasztelnik:
We need `ffaker` in version at least 2.0.0. This dependency need to be installed in dependent project to use atmosphere factories.

/cc @dice-cyfronet/atmo-dev-team please take a look.

Fixes #153

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -40,7 +40,7 @@ group :test do
   gem "generator_spec"
 
   # gem 'factory_girl'
-  gem 'ffaker'
+  gem 'ffaker', '~>2.0.0'
   gem 'database_cleaner'
 
   gem "codeclimate-test-reporter", require: nil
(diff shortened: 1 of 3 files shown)


Nuanda:
:+1:

nowakowski:
:+1:

PR #197, opened 2015-06-30
Simple multitenancy

tbartynski:
TODOs:
- [x] @mkasztelnik : Start Atmo locally and test basic functionality such as starting and stopping a vm. Notes about it here https://redmine.dev.cyfronet.pl/projects/vph-share/wiki/Multitenancy_
- [x] @mkasztelnik: Ensure API v1 compatibility
- [x] Add network_id and tenant_id to Tenant
- [ ] **In Plgrid project** add a periodic task that queries for plg groups that has network id associated and based on this updates information about tenants in the Atmo DB. It requires querying LDAP or Baazar or both. It also need to assign a default fund to a tenant.
- [x] @tbartynski: Default network id provider should return net id from tenant model if tenant.network_id.present? or nil
- [x] @mkasztelnik Make sure that monitoring workers are working correctly.
- [ ] Make sure that for a shared tmpl OpenStack returns image object with one id_at_site.
- [ ]  Temporarily change VMT and Tenant relation to has_and_belongs_to_many since one tmpl maybe shared among tenants.
- [ ]  If Atomic Service is available for more than one tenant user should be given a choice of tenants that can access this tmpl and that are available for him (based on plg groups).
- [ ] There is a many-to-many relation between tenants and tmpls therefore if monitoring running in context of one tenant cannot see a tmpl any longer it should remove the relation with tenant and destroy the tmpl only if tmpl is not associated with ayny other tenant.

Issues:
- [ ] What should we do if Tenant disappears and it used to have vms or tmpls that are still  in cloud? If tenant disappears from OpenStack also Atmo cannot do anything since it lost credentials for this specific tenant in cloud.


mkasztelnik:
I'm dealing with API v1 compatibility right now

mkasztelnik:
I propose to validate if refactoring from CS into Tenant is ok (manual testinng of UI, basic appliance lifecycle), API v1 is compatible and merge it ASAP. After that in short living branches adds features described by @tbartynski in this PR.

mkasztelnik:
We can do it similar like I done it with plgapp security audit (https://github.com/dice-cyfronet/plgapp/issues/79): one issue with todo tasks and a refference to PR which solves concrete problem. What do you think?

mkasztelnik:
Just to remember/check:

I believe this change is incorrect:

```
def set_compute_sites
-    @compute_sites = Atmosphere::ComputeSite.all
+    @tenants = Atmosphere::Tenant.all
end
```

most probably it will blow up admin virtual machine templates index view. This method is only used there, so we should move it from general `ApplicationController`.

nowakowski:
Re: TODO#3: there already exists a column named `tenant_id` in `Tenant` (it's the old `compute_site_id` from `ComputeSite`). Do we need another specialized column?

nowakowski:
@mkasztelnik I'm okay with the proposed testing methodology. Can you or TB deploy the current branch to plgrid-dev (I assume that's where you want to test it)?

nowakowski:
Btw., views may blow up since they're poorly covered by rspecs. We'll deal with issues as they emerge.

mkasztelnik:
@nowakowski yes we need both `site_id` and `tenant_id`. The first one will be used by redirus to generate adding/removing/editing http/https request into appropriate queue and the `tenant_id` will be used to synchronize concrete tenant with ldap/bazar.

@nowakowski do you have knowledge how ldap schema for tenants and groups will look like?

mkasztelnik:
We definitely need to fix hound issues since it generates a lot of noise.

nowakowski:
@mkasztelnik I pushed the requested change related to `set_compute_sites` in `ApplicationController` but I propose we avoid getting sidetracked in this particular branch and focus on what's necessary. The job is big enough as it is.

nowakowski:
@mkasztelnik Re: `tenant_id` - I propose we simply use the existing field for the purposes mentioned in your comment. I have added `network_id` although from what JM has told us it will no longer be necessary to add network ID information to cloud_site requests when spawning VMs for a particular tenant.

I have no knowledge regarding the tenant representation in the LDAP, however the agreement with MR is that the name of each tenant will match the name of the PL-Grid team exactly (so `plgg-cloud` etc.) Can you (or @tbartynski ) tell us what information is required in order to launch a VM in the context of a specific OST tenant?

tbartynski:
@nowakowski I've added site_id field to Tenant since we need it for Redirus. We need a common id for all tenants.
Regarding getting information about tenants I'm not sure yet but will try to figure it out today.

nowakowski:
Okay; I'm going to stop trying to appease Hound for now since the rabbit hole goes very deep and I seem to be just making things worse. :angry:

Making the entire Atmosphere project Hound-compliant will be a massive undertaking and it certainly shouldn't be taking place in this PR. Sorry for the mail spam. :)

nowakowski:
@tbartynski  `site_id` belongs in `ComputeSite` (and you can have a delegation in `Tenant` to avoid painful refactoring).

nowakowski:
@tbartynski To clarify - I would propose:
- Recreating `ComputeSite` with a 1:n relationship with `Tenant`
- Adding `site_id` to `ComputeSite`
- Delegating `Tenant.site_id` to parent.

Much more elegant than cramming `site_id` into `Tenant` (IMO :)

@mkasztelnik Your opinion...?

tbartynski:
@nowakowski I also like your suggestion better in terms of elegance. Please remember we've decided to keep things as simple as possible and not to introduce CS model again at this stage. I am aware it will be required in near future anyway. It's your decision guys since you'll working with that in upcoming days.

nowakowski:
It seems to me that at some point *not* introducing `ComputeSite` does not make things any simpler -- quite the opposite in fact. I would view this circumstance as an example. Still, let @mkasztelnik decide; he's a better programmer than myself. :)

tbartynski:
By the way, I've created a page on redmine wiki https://redmine.dev.cyfronet.pl/projects/vph-share/wiki/Multitenancy_
Please indicate there what works fine and what does not. I hope this help us to organize our work.

mkasztelnik:
@tbartynski, @nowakowski lets finish for now Tenant revolution and leave everything else as it is. According to `site_id` we have strong argument to add it. I believe that f2f meeting on Monday will allow us to discuss it deeply. 

The goal for this week would be to have atmosphere installation bulletproof (as bulletproof as I was before Tenant refactoring) - so existing clew should be able to start new appliance, manage it (http/https redirections, tcp/udp redirections, saving as AT), all Admin views are working, all delayed jobs are working correctly. After this we will focus on automatic tenant generation/removal.

Tomorrow in the morning I should be able to push API v1 compliance changes and then I will focus on testing and correcting admin views.

nowakowski:
I'll work on the `VirtualMachineTemplate` <> `Tenant` relationship.

mkasztelnik:
@tbartynski your last changes introduced a lot of spec failures. Are you able to fix it before your holiday?

@nowakowski we need to be sure that current version is working correctly before introducing new changes. Can you pull current changes, pull database from vph-dev (or create separate CS on your local machine) and perform detailed manual tests? I will also try to do it today and verify if everything is working as expected.

nowakowski:
@mkasztelnik This is being done for PL-Grid; perhaps it would make more sense to import the cloud-dev database rather than the vph-dev database?

mkasztelnik:
@nowakowski vph-dev cloud is more reliable than cloud-dev (ask @mpawlik for details :smile:), but of course you can try with plg cloud-dev if you want.

nowakowski:
@mkasztelnik @mpawlik 's beef was with the production PL-Grid infrastructure, not with cloud-dev.

nowakowski:
I have verified that OpenStack returns identical templates on multiple tenants -- hence the relation between `VirtualMachineTemplate` and `Tenant` should indeed be m:n. I will continue to hack away at it.

mkasztelnik:
I propose to finish tenant stabilization in this PR and don't add additional things here => m:n between VMT and Tenant new branch.

nowakowski:
@mkasztelnik It's already done (i.e. the branching, not the m:n).

mkasztelnik:
All problems with UI fixed, API v1 compatibility guaranteed, deployed into plg cloud-dev. IMHO we should merge it and work for further extension in short living branches.

PR #203, opened 2015-07-08
Change VMT+Flavor selection logic to acknowledge Tenant-User relationship through Funds

nowakowski:
The Optimizer has been altered in such a way that it only preselects those Virtual Machine Templates which belong to Tenants accessible to the Appliance's owner (i.e. those which share at least one Fund with the user who owns the Appliance's ApplianceSet).

This change broke a lot of old RSpecs due to the way factories operate (creating superfluous User and Tenant objects which are not bound to one another). Fixes were applied wherever required.


nowakowski:
All done as requested. Mind the TODOs. @mkasztelnik could you take another look and merge if OK?

app/models/atmosphere/tenant.rb
```diff
@@ -124,6 +124,8 @@ class Tenant < ActiveRecord::Base
 
     scope :active, -> { where(active: true) }
 
+    scope :inactive, -> { where(active: false) }
```
mkasztelnik: This is not used in the code - should be removed.

app/models/atmosphere/user.rb
```diff
@@ -103,6 +103,11 @@ def clew_roles
       roles.map { |r| r == :admin ? 'cloudadmin' : r.to_s }
     end
 
+    # Returns a list of this user's tenants to which the user is linked via fund assignments
+    def tenants
+      funds.map(&:tenants).flatten.uniq.compact
```
mkasztelnik: I see `n+1` issue here. Please use arel query instead

app/providers/atmosphere/optimization_strategy/default.rb
```diff
@@ -35,17 +35,39 @@ def can_scale_manually?
...
       private
 
       attr_reader :appliance
 
+      def restrict_by_user_requirements(vmts, appliance)
+        # If the user requests that the appliance be bound to a specific set of tenants,
+        # the optimizer should honor this selection.
+
+        # Workaround
+        funded_active_tenants = appliance.tenants.funded_by(appliance.fund).reject{|t| t.active == false}
```
mkasztelnik: Why arel query is not used here instead of traversing though collection?

app/providers/atmosphere/optimization_strategy/default.rb
```diff
@@ -35,17 +35,39 @@ def can_scale_manually?
...
 
       private
 
       attr_reader :appliance
 
+      def restrict_by_user_requirements(vmts, appliance)
+        # If the user requests that the appliance be bound to a specific set of tenants,
+        # the optimizer should honor this selection.
+
+        # Workaround
```
mkasztelnik: I'm not sure if I understand this comment. Can you explain why this is workaround?

app/providers/atmosphere/optimization_strategy/default.rb
```diff
@@ -37,15 +37,33 @@ def vmt_candidates_for(appliance)
...
+        vmts = restrict_by_tenant_availability(vmts, appliance)
+        vmts
       end
 
       private
 
       attr_reader :appliance
 
+      def restrict_by_user_requirements(vmts, appliance)
+        # If the user requests that the appliance be bound to a specific set of tenants,
```
mkasztelnik: If you want to create method documentation please use rdocs syntax before method instead of adding comment inside method.

app/providers/atmosphere/optimization_strategy/default.rb
```diff
@@ -37,15 +37,33 @@ def vmt_candidates_for(appliance)
...
+        # the optimizer should honor this selection.
+        vmts.select do |vmt|
+          (vmt.tenants & appliance.tenants.active.funded_by(appliance.fund)).present?
+        end
+      end
+
+      def restrict_by_tenant_availability(vmts, appliance)
+        # In all cases the optimizer should only suggest those vmts which the user is able to access
```
mkasztelnik: Similar as above

app/providers/atmosphere/optimization_strategy/default.rb
```diff
@@ -64,41 +86,67 @@ def preferences
...
-                end
-              ) {|f| tmpl.get_hourly_cost_for(f) || Float::INFINITY }
-            ).sort!{ |x,y| y.memory <=> x.memory }.last
-            hsh[opt_fl] = tmpl if opt_fl
-            hsh
+          # We are given a list of tmpls which MAY be spawned by the user in the context of @appliance.
+          # We are asked to return a single template, a specific flavor and a specific tenant for instantiation.
+          # Flavor prices may vary by tenant.
+          # !CAUTION! @appliance can be nil -- this will happen when the method is invoked prior to
+          # creation of an appliance.
```
mkasztelnik: Please use rdoc before method instead of comments inside method

app/providers/atmosphere/optimization_strategy/default.rb
```diff
@@ -64,41 +86,67 @@ def preferences
...
+          best_flavor = nil
+          instantiation_cost = Float::INFINITY
+
+          tmpls.each do |tmpl|
+            # Assuming tmpl is selected, figure out which tenant and which flavor to use.
+            candidate_tenants = get_candidate_tenants_for_template(tmpl)
+            # For each candidate tenant find optimal flavor and the corresponding VM instantiation cost
+            # Note that candidate_tenants can be blank - we must handle this case.
+            opt_tenant = nil
+            opt_flavor_and_cost = [nil, Float::INFINITY]
```
mkasztelnik: Maybe hash or struct with meaningful keys would be more understandable than `opt_flavor_and_cost[1]`?

E.g.

```ruby
opt_flavor_and_cost = { flavor: nil, price: Float::INFINITY }
# or
opt_flavor_and_cost = Struct.new(:flavor, :price).new(nil, Float::INFINITY)
```

and then:

```ruby
opt_flavor_and_cost[:flavor]
opt_flavor_and_cost[:price]
#or
opt_flavor_and_cost.flavor
opt_flavor_and_cost.price
```

mkasztelnik:
To understand whole default optimization logic I need fresh brain - will continue tomorrow morning.

What about hound errors?

app/providers/atmosphere/optimization_strategy/default.rb
```diff
@@ -35,17 +35,39 @@ def can_scale_manually?
...
 
+      def restrict_by_user_requirements(vmts, appliance)
+        # If the user requests that the appliance be bound to a specific set of tenants,
+        # the optimizer should honor this selection.
+
+        # Workaround
+        funded_active_tenants = appliance.tenants.funded_by(appliance.fund).reject{|t| t.active == false}
+
+        vmts.select do |vmt|
+          (vmt.tenants & funded_active_tenants).present?
```
mkasztelnik: If arel query would be used in line 56 than I believe we can have (potentially - should be checked because I'm not 100% sure) single sql query (with subquery): 
```ruby
funded_active_tenants.where(id: vmt.tenants.pluck(:id)).count > 0
```

app/providers/atmosphere/optimization_strategy/default.rb
```diff
@@ -35,17 +35,39 @@ def can_scale_manually?
...
+        end
+      end
+
+      def restrict_by_tenant_availability(vmts, appliance)
+        # In all cases the optimizer should only suggest those vmts which the user is able to access
+        # (i.e. vmts which reside on at least one tenant which shares a fund with the appliance).
+        vmts.select do |vmt|
+          (vmt.tenants & appliance.fund.tenants).present?
```
mkasztelnik: Similar situation as in previous method. This is not critical issue, but wart to check :)

app/providers/atmosphere/optimization_strategy/default.rb
```diff
@@ -64,41 +86,67 @@ def preferences
         end
       end
 
-      class VmAndFlavor
+      class VmAndFlavorAndTenant
```
mkasztelnik: Most probably here was misspell from the beginning: Vmt instead of Vm

app/providers/atmosphere/optimization_strategy/default.rb
```diff
@@ -35,17 +35,39 @@ def can_scale_manually?
...
-          state: 'active',
+          state: 'active'
         )
-        vmts.select do |vmt|
-          (vmt.tenants & appliance.tenants.active.funded_by(appliance.fund)).present?
+        if appliance.tenants.present?
+          vmts = restrict_by_user_requirements(vmts, appliance)
         end
+        vmts = restrict_by_tenant_availability(vmts, appliance)
+        vmts
```
mkasztelnik: Usage of variable is not needed here use simply:

```ruby
restrict_by_tenant_availability(vmts, appliance)
```

app/providers/atmosphere/optimization_strategy/default.rb
```diff
@@ -64,41 +86,67 @@ def preferences
...
+            candidate_tenants.each do |t|
+              opt_flavor, cost = get_optimal_flavor_for_tenant(tmpl, t)
+              if cost < opt_flavor_and_cost[1]
+                opt_tenant = t
+                opt_flavor_and_cost = [opt_flavor, cost]
+              end
+            end
+            # An optimal tenant may not have been found - in that case skip tmpl entirely.
+            # Otherwise check if the tenant that has been found is better than the current "leader"
+            if opt_tenant.present?
```
mkasztelnik: Why we have 2 checks: first one checking for the best flavor in the `vmt` scope and next checking the best flavor? Would it be enough to only check if flavor is "the best" since we are only returning one value?

app/providers/atmosphere/optimization_strategy/default.rb
```diff
@@ -64,41 +86,67 @@ def preferences
...
-              opt_flavors_and_tmpls_map.keys
-            ) { |f| tmpls.first.get_hourly_cost_for(f)}
-          ).sort { |x,y| x.memory <=> y.memory }.last
+          # It is possible that nothing has been found - this would indicate unsatisfiable hardware requirements
+          # (i.e. no flavor with sufficient CPU/memory/HDD) or non-existence of flavors supporting the selected
+          # architecture.
+          # If this occurs, return first template to avoid breaking user notification logic
+          # in Atmosphere::Cloud::SatisfyAppliance
+          # TODO: This is quite hacky; a better way to communicate errors to the user is needed.
+          if best_template.blank?
```
mkasztelnik: Is this hack allow to start appliance in the tenant user does not belongs to or no?

app/providers/atmosphere/optimization_strategy/default.rb
```diff
@@ -64,41 +86,67 @@ def preferences
...
+            best_template = tmpls.first
+          end
 
           [
```
mkasztelnik: I know that it is not in the scope of this PR but maybe you remember why we are returning array here instead of single the best value?

app/providers/atmosphere/optimization_strategy/default.rb
```diff
@@ -107,6 +155,49 @@ def select
...
+        def get_candidate_tenants_for_template(tmpl)
+          # Determine which tenants can be used to spawn this specific template in the context
+          # of the current @appliance.
+          # If @appliance is nil (which may happen) then return all of this tmpl's tenants:
+          if @appliance.blank?
+            # This will happen in pre-instantiation queries (such as in the CLEW
+            # "Start application" window) where no appliance is yet present.
+            tmpl.tenants
+          else
+            if @appliance.tenants.present?
```
mkasztelnik: Please extract shared `if` `else` elements before this condition:

```
tenants = tmpl.tenants & @appliance.fund.tenants & @appliance.appliance_set.user.tenants
tenants &= @appliance.tenants if @appliance.tenants.present?
```

spec/services/atmosphere/cloud/satisfy_appliance_spec.rb
```diff
@@ -9,16 +9,17 @@
...
   end
 
-  let!(:wf) { create(:workflow_appliance_set) }
-  let!(:wf2) { create(:workflow_appliance_set) }
-  let!(:shareable_appl_type) { create(:shareable_appliance_type) }
   let!(:fund) { create(:fund) }
+  let!(:u) { create(:user, funds: [fund]) }
+  let!(:wf) { create(:workflow_appliance_set, user: u) }
+  let!(:wf2) { create(:workflow_appliance_set, user: u) }
+  let!(:shareable_appl_type) { create(:shareable_appliance_type) }
```
mkasztelnik: 4 new objects will be created before each spec. Are you sure these 4 elements are needed in every test in this file?

spec/services/atmosphere/optimizer_spec.rb
```diff
@@ -7,18 +7,19 @@
...
 
-  let!(:wf) { create(:workflow_appliance_set) }
-  let!(:wf2) { create(:workflow_appliance_set) }
-  let!(:shareable_appl_type) { create(:shareable_appliance_type) }
   let!(:fund) { create(:fund) }
   let!(:openstack) { create(:openstack_with_flavors, funds: [fund]) }
+  let!(:u) { create(:user, funds: [fund]) }
+  let!(:wf) { create(:workflow_appliance_set, user: u) }
+  let!(:wf2) { create(:workflow_appliance_set, user: u) }
+  let!(:shareable_appl_type) { create(:shareable_appliance_type) }
```
mkasztelnik: Similar as above, are these objects used in every test in this file?

app/models/atmosphere/tenant.rb
nowakowski: OK.

app/models/atmosphere/user.rb
nowakowski: Okay.

app/providers/atmosphere/optimization_strategy/default.rb
nowakowski: Yep.

app/providers/atmosphere/optimization_strategy/default.rb
nowakowski: Affirmative.

app/providers/atmosphere/optimization_strategy/default.rb
nowakowski: Old code left behind from my experiments with `Tenant.active`. Sorry. Fixed.

app/providers/atmosphere/optimization_strategy/default.rb
nowakowski: Had trouble with `Tenants.active`. Works now. Fixed.

app/providers/atmosphere/optimization_strategy/default.rb
nowakowski: Still 2 queries. I believe it's because of the `select` block. Leaving as is since this is not a performance bottleneck (only 1 extra query and the method is only invoked while launching an `Appliance`).

app/providers/atmosphere/optimization_strategy/default.rb
nowakowski: OK, fixing.

app/providers/atmosphere/optimization_strategy/default.rb
nowakowski: Yep.

app/providers/atmosphere/optimization_strategy/default.rb
nowakowski: Got rid of this block altogether.

app/providers/atmosphere/optimization_strategy/default.rb
nowakowski: Yeah, sloppy. Fixed. Thanks.

app/providers/atmosphere/optimization_strategy/default.rb
mkasztelnik: Puck was needed in my example. Try something like:

```
[6] pry(main)> Atmosphere::Tenant.where(id: Atmosphere::Tenant.all).to_sql
=> "SELECT \"atmosphere_tenants\".* FROM \"atmosphere_tenants\" WHERE \"atmosphere_tenants\".\"id\" IN (SELECT \"atmosphere_tenants\".\"id\" FROM \"atmosphere_tenants\")"
```

app/providers/atmosphere/optimization_strategy/default.rb
nowakowski: No, it doesn't. For VM instantiation you need all three items (tmpl, tenant, flavor). Look at `app/services/atmosphere/cloud/satisfy_appliance.rb:23` to find out why this hack is "needed". (Well, it's not actually required, but I don't have time to rewrite that class atm. ;)

app/providers/atmosphere/optimization_strategy/default.rb
nowakowski: Haven't the slightest. Probably a leftover from some of @tbartynski 's early experiments with optimization.

app/providers/atmosphere/optimization_strategy/default.rb
nowakowski: OK.

spec/services/atmosphere/cloud/satisfy_appliance_spec.rb
nowakowski: What four objects? I only see one extra object (`u`). Actually, this change results in **fewer** objects being created due to the way factories work (e.g. if you don't specify `user: u` in `ApplianceSet` creators then FactoryGirl will happily create an entirely new `User` object for each `ApplianceSet` etc.

Oh, and by the way - skipping `u` causes most of the tests in this spec to fail because the "default" users which FG will create are not bound to the correct funds (and therefore fail the new optimizer checks).

spec/services/atmosphere/optimizer_spec.rb
nowakowski: See above. :)

app/providers/atmosphere/optimization_strategy/default.rb
nowakowski: OK, I take it we cannot keep the existing code.

(...1 hour later...)

I think this *probably* does what you want:

<pre>
Atmosphere::VirtualMachineTemplate.joins(:tenants).
where(Atmosphere::VirtualMachineTemplate.arel_table[:id].in(vmts.pluck(:id)).
and(Atmosphere::Tenant.arel_table[:id].in(funded_active_tenants.pluck(:id))))
</pre>

Of course if you know of a better way to write this query (as opposed to "something like this query" ;)) then go ahead and tell me.

God I hate Arel... :angry:

app/providers/atmosphere/optimization_strategy/default.rb
nowakowski: Fixed that too. For a given value of "fixed". ;)

nowakowski:
Well damn. It's that transient again. And I thought I had it fixed... :(

PR #204, opened 2015-07-10
Don't allow to modify fund name after it is created

mkasztelnik:
Fund name is used in plgrid to distinguish default team fund. That is why there should not be possible to edit fund name. I believe that is does not have any drawbacks in other projects.


mkasztelnik:
@nowakowski please take a look

nowakowski:
There's nothing wrong with the code but you're stealthily inserting a PLG-specific feature into Atmosphere core (actually it's not even a feature -- it's a restriction, and it's only here because PLG needs it). If you believe this is OK then :+1:

PR #205, opened 2015-07-10
Use tenant `site_id` to generate redirus tasks

mkasztelnik:
Site is can be shared among many tenants. It is used distinguish correct Redis queue name to generate add/remove http/https proxies jobs.

### DIFF
--- a/spec/factories/tenants.rb
+++ b/spec/factories/tenants.rb
@@ -2,7 +2,7 @@ FactoryGirl.define do
   factory :tenant, aliases: [:openstack_tenant], class: 'Atmosphere::Tenant' do |f|
     tenant_type 'private'
     tenant_id { SecureRandom.hex(4) }
-    site_id { 'somesite' }
+    site_id { "site_id_#{SecureRandom.hex(4)}" }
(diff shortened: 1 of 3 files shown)


spec/factories/tenants.rb
```diff
@@ -2,7 +2,7 @@
   factory :tenant, aliases: [:openstack_tenant], class: 'Atmosphere::Tenant' do |f|
     tenant_type 'private'
     tenant_id { SecureRandom.hex(4) }
-    site_id { 'somesite' }
+    site_id { "site_id_#{SecureRandom.hex(4)}" }
```
nowakowski: I did this on purpose to "simulate" PLGrid-like behavior. Future tests will need to be careful to override this to produce meaningful results (the random assignment is more "VPH-like").

nowakowski:
:+1:

spec/factories/tenants.rb
mkasztelnik: This is a factory, random values are good here. If needed you can assign required value while creating object.

nowakowski: I know, Terrance, I know. :grinning:

PR #209, opened 2015-07-14
Fix fund selection for appliance.

nowakowski:
Updates appliance fund selection logic to acknowledge the contents of appliance.tenants and also the list of funds available to the appliance's owner.

### DIFF
--- a/app/models/atmosphere/appliance.rb
+++ b/app/models/atmosphere/appliance.rb
@@ -163,6 +163,12 @@ module Atmosphere
         vmt.tenants.map(&:funds)
       end.flatten.uniq.compact
 
+      # Restrict to funds which belong to this appliance's owner
+      funds = funds & appliance_set.user.funds
+
+      # Restrict to funds which match this appliance's set of selected tenants
+      funds = funds & tenants.map(&:funds) if tenants.present?
+
       self.fund = if funds.include? default_fund
                     default_fund
                   else
--- a/app/models/atmosphere/appliance.rb
+++ b/app/models/atmosphere/appliance.rb
@@ -156,18 +156,22 @@ module Atmosphere
 
     private
 
+    # TODO: This could be buggy. Make sure to verify that Atmosphere uses
+    # the correct fund to spawn VMs in the selected tenant.
     def assign_fund
-      # TODO: This could be buggy. Make sure to verify that Atmosphere uses the correct fund to spawn VMs
-      # in the selected tenant.
-      funds = appliance_type.virtual_machine_templates.map do |vmt|
-        vmt.tenants.map(&:funds)
-      end.flatten.uniq.compact
+      funds = Fund.joins(tenants: :virtual_machine_templates).
+              where(id: appliance_set.user.funds,
+                    atmosphere_virtual_machine_templates: {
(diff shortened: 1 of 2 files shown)


app/models/atmosphere/appliance.rb
```diff
@@ -163,6 +163,12 @@ def assign_fund
         vmt.tenants.map(&:funds)
       end.flatten.uniq.compact
 
+      # Restrict to funds which belong to this appliance's owner
+      funds = funds & appliance_set.user.funds
+
+      # Restrict to funds which match this appliance's set of selected tenants
+      funds = funds & tenants.map(&:funds) if tenants.present?
```
mkasztelnik: n+1 queries, try something like: 

```ruby
Atmosphere::Fund.joins(:tenants).where(id: funds, tenants: { id: tenants })
```

app/models/atmosphere/appliance.rb
```diff
@@ -163,6 +163,12 @@ def assign_fund
         vmt.tenants.map(&:funds)
       end.flatten.uniq.compact
 
+      # Restrict to funds which belong to this appliance's owner
+      funds = funds & appliance_set.user.funds
```
mkasztelnik: Maybe something like:

```ruby
funds.where(id: appliance_set.user.funds)
```

then most probably there will be 1 sql query instead of 2

mkasztelnik:
Removing ready for review since tests are failing

nowakowski:
@mkasztelnik I know this will feel suspicious but I am deleting the failing test for two reasons:
- The test is bogus. It creates an `Appliance` object, then creates some additional tenants and VMTs and expects the appliance's fund to change. An appliance's fund is determined upon creation and there is no code in Atmosphere which would automatically swap it for another fund.
- My newly added test ('does not assign a fund which is incompatible with selected tenants') checks for exactly the same behavior (except it's correctly written). ;)

app/models/atmosphere/appliance.rb
nowakowski: `funds` is not an AR relation. Sorry, it's late. If you can rewrite the query in lines 162-164 in pure arel (I tried some stuff but am unable to reproduce the required behavior)  then I'll introduce this change; otherwise I believe any potential performance gain is absolutely trivial and not worth the effort.

app/models/atmosphere/appliance.rb
nowakowski: See above.

mkasztelnik:
Unfortunately introduced spec is false positive :( I made small change:
  - user is assigned to `f1` and f2` 
  - 2 appliances are created. First one with `t1` assigned second one with `f2`

For the second one wrong fund is assigned. I will work on this method, but it is once again hack. I think that we should redesign fund assignment logic after releasing rpm for plgrid, because `assign_fund` is limiting possible optimizer choices. The correct solution IMHO is to respect fund assignment when user selects it, in other case let optimizer to set this value after most optimal tenant to start VM is selected.

mkasztelnik:
@nowakowski now you take a look :smile_cat:

nowakowski:
@mkasztelnik Yup, this looks good. Merge if no further comments.

PR #211, opened 2015-07-15
Fix erroneous deletion rule in appliance->tenant relationship

nowakowski:


nowakowski:
@mkasztelnik I have not been able to determine who introduced this bug. It is present in the very first commit of atmosphere/core on github and I cannot find the old atmosphere project on our gitlab. Regardless, the deletion rule is clearly wrong; destroying an appliance should not cause destruction of a tenant (this bug was benign because other validations prevent the tenant from actually being destroyed).

PR #212, opened 2015-07-16
Remove user keys only from active tenants

### LINKED ISSUE #193: User key should not be removed from inactive compute site
Scenario:

  * configure 2 compute sites
  * add user key into both compute sites
  * disable one compute site and change compute site password into invalid
  * try to remove user key

As a result 500 error code is returned.

mkasztelnik:
Fix #193


mkasztelnik:

nowakowski:
:+1:

PR #217, opened 2015-08-06
Move optimize appliance into service from callback

mkasztelnik:
This is a first step to make optimization asynchronous. Optimization execution was removed from appliance callback and moved into dedicated service.


mkasztelnik:

tbartynski:
:+1:

PR #218, opened 2015-08-07
Asynchronous appliance optimization

mkasztelnik:
After appliance is create delayed job is scheduled to optimize it. This is working from atmosphere point of view. Unfortunately CLEW is not ready for this change. It assumes that at least on VM will be created after appliance is saved. As a result "unexpected error occurred" message is shown to the user (after refreshing CLEW new appliance is shown correctly). @dharezlak will be looking at this issue in next week (you can track this issue here https://github.com/dice-cyfronet/clew/issues/12). 

@dharezlak after  https://github.com/dice-cyfronet/clew/issues/12 is solved please add compiled CLEW version into `public/clew` in this branch, then we will merge it into master.

EDIT:
This branch is deployed into vph-dev where you can test it.


mkasztelnik:

PR #224, opened 2015-08-17
Fix openstack vm destroy in fog 1.33

mkasztelnik:
Following error was thrown:

```
Fog::OpenStack::Errors::InterfaceNotImplemented:
Method :destroy is not implemented
```


tbartynski:
:+1:

PR #227, opened 2015-08-17
Fix user key form

### LINKED ISSUE #164: New user key form rendered with error after an attemp to add DSA key
DSA keys are not supported. If one tries to add a DSA key an appropriate error message is displayed and form for adding a new key is shown with all fields filled with content. User selector widget is broken a bit however. It displays DB ids of users instead of their logins. It displays login fine for the first time.

mkasztelnik:
Use simple form association for selecting user key owner

Fix #164


tbartynski:
:+1:

PR #228, opened 2015-08-18
Allow delete non managed VM from Atmosphere DB

### LINKED ISSUE #225: Monitoring is not able to remove VM started outside atmosphere
Scenario:
  * start VM using e.g. nova client (`managed_by_atmosphere` should be set to false in atmosphere VM)
  * remove VM using e.g. nova client

as a result VM cannot be removed from atmosphere because of https://github.com/dice-cyfronet/atmosphere/blob/master/app/models/atmosphere/virtual_machine.rb#L233-L238

mkasztelnik:
Fix #225 

/cc @dice-cyfronet/atmo-dev-team please take a look

### DIFF
--- a/app/models/atmosphere/virtual_machine.rb
+++ b/app/models/atmosphere/virtual_machine.rb
@@ -67,7 +67,6 @@ module Atmosphere
 
     before_update :update_in_monitoring, if: :ip_changed?
     before_destroy :unregister_from_monitoring, if: :in_monitoring?
-    before_destroy :cant_destroy_non_managed_vm
     after_update :regenerate_dnat, if: :ip_changed?
     after_destroy :delete_dnat, if: :ip?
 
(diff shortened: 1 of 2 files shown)


tbartynski:
The code looks fine, however I guess there used to be such a guard for a reason. I can't think of it now. Since we check if vm is managed by Atmosphere in VM model before we actually delete it in cloud your modification seems safe, however I would recommend to have someone else to look at this PR also.
 Shy :+1:

mkasztelnik:
IMHO this is safe. When we delete VM which is not managed by atmosphere it is deleted only from Atmosphere DB, not from the cloud. So during next VM monitoring worker it will be recreated :)

mkasztelnik:
This PR should be also backported into stable branch.

nowakowski:
:+1:

mkasztelnik:
also cherry-picket into stable branch

PR #235, opened 2015-08-20
Add validation for max fund balance value

### LINKED ISSUE #223: Entering large fund values causes http500 errrors
When supplying a large value (exceeding psql 4b int range) in "Balance" field on "Editing Fund", the server responds with http500 error. This basically limits funds value to 214748.3647 of currency. This might limit scalability of our system.
pb

mkasztelnik:
Fix #223

@dice-cyfronet/atmo-dev-team, @mpawlik please take a look

### DIFF
--- a/app/models/atmosphere/fund.rb
+++ b/app/models/atmosphere/fund.rb
@@ -48,7 +48,10 @@ module Atmosphere
               presence: true
 
     validates :balance,
-              numericality: true
+              numericality: {
+                only_integer: true,
+                less_than_or_equal_to: 2147483647
+              }
 
     validates :overdraft_limit,
               numericality: { less_than_or_equal_to: 0 }


app/models/atmosphere/fund.rb
```diff
@@ -48,7 +48,10 @@ class Fund < ActiveRecord::Base
               presence: true
 
     validates :balance,
-              numericality: true
+              numericality: {
+                only_integer: true,
+                less_than_or_equal_to: 2147483647
```
nowakowski: Huh...? :astonished:

mkasztelnik: Maximum value for an integer, which is used for this field in postgresql.

mkasztelnik: I can convert it into double, but right now I believe that 2147483647 is big enough :smile:

nowakowski: Oh, okay. Who knew? :)

nowakowski:
:+1:

app/models/atmosphere/fund.rb
mkasztelnik: http://www.postgresql.org/docs/9.1/static/datatype-numeric.html :)

PR #238, opened 2015-08-21
Use tenant_id when site_id is nil

### LINKED ISSUE #237: All add/remove http mappings in incorrect queue
on vph-dev all add/remove http mappings go into `default` queue instead on queue equals to `tenant_id`. This bug can be connected with sidekiq update needed because of critical security fix.

mkasztelnik:
This change guarantees backward compatibility. `site_id` is used to get
correct queue name while generating job to add/remove http/https
redirection. Previously `tenant_id` was used for it. This change
guarantee backward compatibility and allow to override queue using
`site_id` field.

Fix #237 

@dice-cyfronet/atmo-dev-team please take a look.

### DIFF
--- a/app/models/atmosphere/tenant.rb
+++ b/app/models/atmosphere/tenant.rb
@@ -181,6 +181,10 @@ module Atmosphere
         where("atmosphere_tenants.site_id = '#{site_id}'").all
     end
 
+    def site_id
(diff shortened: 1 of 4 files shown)


nowakowski:
:+1:

tbartynski:
:+1:

PR #241, opened 2015-08-25
Fix delete vm vmt

mkasztelnik:
This PR restores old fog implementation, where were possible to remove server and image using `servers` and `images` model.


mkasztelnik:
I tested delete on openstack an amazon - works ok. 

tbartynski:
:+1:

PR #242, opened 2015-08-25
Updated fund assignment logic

### LINKED ISSUE #210: Rework fund assignment to appliances
We should redesign fund assignment logic after releasing rpm for plgrid, because assign_fund is limiting possible optimizer choices. The correct solution IMHO is to respect fund assignment when user selects it, in other case let optimizer to set this value after most optimal tenant to start VM is selected.

nowakowski:
Fixes #210

### DIFF
--- a/app/models/atmosphere/appliance.rb
+++ b/app/models/atmosphere/appliance.rb
@@ -60,7 +60,6 @@ module Atmosphere
     attr_readonly :dev_mode_property_set
 
     before_create :create_dev_mode_property_set, if: :development?
-    #before_save :assign_fund, if: 'fund.nil?'
     after_destroy :remove_appliance_configuration_instance_if_needed
 
     scope :started_on_site, ->(tenant) do
--- a/app/models/atmosphere/appliance.rb
+++ b/app/models/atmosphere/appliance.rb
@@ -137,30 +137,6 @@ module Atmosphere
 
     private
 
-    # TODO: This could be buggy. Make sure to verify that Atmosphere uses
-    # the correct fund to spawn VMs in the selected tenant.
-    def assign_fund
-      funds = Fund.joins(tenants: :virtual_machine_templates).
-              where(id: appliance_set.user.funds,
-                    atmosphere_virtual_machine_templates: {
-                      appliance_type_id: appliance_type.id
-                    })
-
-      if tenants.present?
-        # we need to use map here, because this code can be invoked
-        # before first appliance save. Then appliance does not have id
-        # assigned which is required to perform tenants query with appliance
-        # join.
-        funds = funds.where(atmosphere_tenants: { id: tenants.map(&:id) })
-      end
-
-      self.fund = if funds.include? default_fund
-                    default_fund
-                  else
-                    funds.blank? ? default_fund : funds.first
-                  end
-    end
-
     def default_fund
       if appliance_set && appliance_set.user
         appliance_set.user.default_fund
--- a/app/models/atmosphere/appliance.rb
+++ b/app/models/atmosphere/appliance.rb
@@ -60,7 +60,7 @@ module Atmosphere
     attr_readonly :dev_mode_property_set
 
     before_create :create_dev_mode_property_set, if: :development?
-    before_save :assign_fund, if: 'fund.nil?'
+    #before_save :assign_fund, if: 'fund.nil?'
     after_destroy :remove_appliance_configuration_instance_if_needed
 
     scope :started_on_site, ->(tenant) do
--- a/app/models/atmosphere/appliance.rb
+++ b/app/models/atmosphere/appliance.rb
@@ -137,12 +137,6 @@ module Atmosphere
 
     private
 
-    def default_fund
-      if appliance_set && appliance_set.user
-        appliance_set.user.default_fund
-      end
-    end
-
     def remove_appliance_configuration_instance_if_needed
       if appliance_configuration_instance &&
          appliance_configuration_instance.appliances.blank?
--- a/app/providers/atmosphere/optimization_strategy/default.rb
+++ b/app/providers/atmosphere/optimization_strategy/default.rb
@@ -100,7 +100,6 @@ module Atmosphere
           best_tenant = nil
           best_flavor = nil
           instantiation_cost = Float::INFINITY
-
           tmpls.each do |tmpl|
             candidate_tenants = get_candidate_tenants_for_template(tmpl)
             candidate_tenants.each do |t|
--- a/app/providers/atmosphere/optimization_strategy/default.rb
+++ b/app/providers/atmosphere/optimization_strategy/default.rb
@@ -162,7 +162,7 @@ module Atmosphere
         def get_candidate_tenants_for_template(tmpl)
           # Determine which tenants can be used to spawn this specific template in the context
           # of the current @appliance.
-          # If @appliance is nil (which may happen) then return all of this tmpl's active tenants:
+          # If @appliance is nil (which may happen) return active tenants:
           if @appliance.blank?
             # This will happen in pre-instantiation queries (such as in the CLEW
             # "Start application" window) where no appliance is yet present.
--- a/app/providers/atmosphere/optimization_strategy/default.rb
+++ b/app/providers/atmosphere/optimization_strategy/default.rb
@@ -41,7 +41,6 @@ module Atmosphere
           vmts = restrict_by_user_requirements(vmts)
         end
         vmts
-        #restrict_by_tenant_availability(vmts)
       end
 
       private
@@ -111,15 +110,15 @@ module Atmosphere
           tmpls.each do |tmpl|
             candidate_tenants = get_candidate_tenants_for_template(tmpl)
             candidate_tenants.each do |t|
-
               # The next step is to restrict tenants by user funds.
               # A tenant is only valid for use if it
               # shares a fund with the appliance's owner. Additionally,
               # if the appliance fund is explicitly specified
               # in the instantiation requests, the selection must be honored.
               unless @appliance.blank?
-                candidate_tenants.select do |t|
-                  candidate_funds = @appliance.appliance_set.user.funds & t.funds
+                candidate_tenants.select do |tenant|
+                  candidate_funds = @appliance.appliance_set.user.funds & \
+                    tenant.funds
                   unless @appliance.fund.blank?
                     candidate_funds = candidate_funds & [@appliance.fund]
                   end
--- a/app/providers/atmosphere/optimization_strategy/default.rb
+++ b/app/providers/atmosphere/optimization_strategy/default.rb
@@ -33,9 +33,10 @@ module Atmosphere
       protected
 
       def vmt_candidates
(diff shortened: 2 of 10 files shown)


mkasztelnik:
@nowakowski you can configure rubymine to validate code using rubocop with hound configuration. Ask @tbartynski how to do it (I can help you to configure rubycop validation in sublimetext). 

Let me know after hound warning will be fixed.

mkasztelnik:
I'm removing `ready for review` tag till hound issues are fixed.

nowakowski:
@mkasztelnik Fixed hound issues and some other bugs which emerged. The rubocop plugin doesn't work for me (throws NPEs), possibly because I'm using an old version of Rubymine.

app/services/atmosphere/cloud/satisfy_appliance.rb
```diff
@@ -13,6 +13,13 @@ def execute
             !(vm = optimization_strategy.vm_to_reuse).nil?
 
             action.log(I18n.t('satisfy_appliance.reusing', vm: vm.id_at_site))
+            # Assign correct fund for this VM to appliance
+            cfs = vm.tenant.funds & appliance.appliance_set.user.funds
+            if cfs.include? appliance.appliance_set.user.default_fund
+              appliance.fund == appliance.appliance_set.user.default_fund
```
mkasztelnik: `==`? it seems like a bug - did you mean `=`?

nowakowski:
@mkasztelnik I just realized that part hadn't been covered by specs. Fixed and added spec.

app/models/atmosphere/appliance.rb
```diff
@@ -60,7 +60,6 @@ class Appliance < ActiveRecord::Base
     attr_readonly :dev_mode_property_set
 
     before_create :create_dev_mode_property_set, if: :development?
-    before_save :assign_fund, if: 'fund.nil?'
```
mkasztelnik: `Atmosphere::Appliance#assign_fund` is not used. This method should be removed.

app/providers/atmosphere/optimization_strategy/default.rb
```diff
@@ -33,14 +33,15 @@ def can_scale_manually?
...
+        vmts = VirtualMachineTemplate.joins(:tenants).where(
           appliance_type: appliance.appliance_type,
-          state: 'active'
+          state: 'active',
+          atmosphere_tenants: { active: true }
         )
         if appliance.tenants.present?
           vmts = restrict_by_user_requirements(vmts)
         end
-        restrict_by_tenant_availability(vmts)
```
mkasztelnik: `restrict_by_tenant_availability` method is not used anymore, it should be removed.

mkasztelnik: `user_selected_tenants` is used only by `restrict_by_tenant_availability`, it should be removed.

app/providers/atmosphere/optimization_strategy/default.rb
```diff
@@ -106,6 +111,21 @@ def select
           tmpls.each do |tmpl|
             candidate_tenants = get_candidate_tenants_for_template(tmpl)
             candidate_tenants.each do |t|
+              # The next step is to restrict tenants by user funds.
+              # A tenant is only valid for use if it
+              # shares a fund with the appliance's owner. Additionally,
+              # if the appliance fund is explicitly specified
+              # in the instantiation requests, the selection must be honored.
+              unless @appliance.blank?
```
mkasztelnik: As far as I understand this guard (and the one in line 166) is needed because we use default policy to get the price for concrete appliance type. Do I understand correctly?

app/services/atmosphere/cloud/satisfy_appliance.rb
```diff
@@ -11,8 +11,16 @@ def execute
...
             action.log(I18n.t('satisfy_appliance.reusing', vm: vm.id_at_site))
+            # Assign correct fund for this VM to appliance
+            cfs = vm.tenant.funds & appliance.appliance_set.user.funds
+            if cfs.include? appliance.appliance_set.user.default_fund
+              appliance.fund = appliance.appliance_set.user.default_fund
+              appliance.save
+            elsif cfs.length > 0
+              appliance.fund = cfs.first
+              appliance.save
+            end
```
mkasztelnik: Please extract this code into separate method (now we have different level of abstraction code in one method). You can use following construction to assign `fund` only when `nil`.

```ruby
appliance.fund ||= select_fund(vm)
```

app/services/atmosphere/cloud/satisfy_appliance.rb
```diff
@@ -39,6 +47,17 @@ def execute
...
+                # Select fund to assign to appliance, if not yet assigned
+                unless appliance.fund.present?
+                  cfs = tenant.funds & appliance.appliance_set.user.funds
+                  if cfs.include? \
+                    appliance.appliance_set.user.default_fund
+                    appliance.fund = appliance.appliance_set.user.default_fund
+                  else
+                    appliance.fund = cfs.first
+                  end
+                end
```
mkasztelnik: Similar as above, extract into separate method and use `||=` to assign fund only when needed.

spec/services/atmosphere/cloud/satisfy_appliance_spec.rb
```diff
@@ -17,6 +17,95 @@
...
+    let!(:appliance) do
+      create(:appliance,
+             appliance_set: appliance_set,
+             appliance_type: appliance_type,
+             fund: nil)
+    end
+
+    it 'does not change fund when externally assigned' do
+      funded_appliance = create(:appliance)
+      expect(funded_appliance.fund).not_to eq appliance.send(:default_fund)
```
mkasztelnik: Don't use private method (what is more after removing `assign_fund` method `default_fund` can also be removed).

mkasztelnik: What is more I don't understand this spec. Now when appliance is created there is no callback which can assign fund, optimization policy is responsible for it.

spec/services/atmosphere/cloud/satisfy_appliance_spec.rb
```diff
@@ -17,6 +17,95 @@
...
+    end
+
+    it 'gets default fund from its user if no fund is set' do
+      appliance.reload
+      Atmosphere::Cloud::SatisfyAppliance.new(appliance).execute
+      expect(appliance.fund).to eq appliance.send(:default_fund)
```
mkasztelnik: Similar as above - don't use private methods only public class API. Otherwise test is fragile and depending on internal class structure. Instead use simply: `user.default_fund`

spec/services/atmosphere/cloud/satisfy_appliance_spec.rb
```diff
@@ -17,6 +17,95 @@
...
+    end
+
+    it 'prefers default fund if it supports relevant tenant' do
+      appliance.reload
```
mkasztelnik: Is it really needed to refresh so many times? Maybe it is enough to refresh appliance once in line 72?

spec/services/atmosphere/cloud/satisfy_appliance_spec.rb
```diff
@@ -17,6 +17,95 @@
...
+      create(:virtual_machine_template,
+             appliance_type: appliance_type,
+             tenants: [default_t])
+      create(:virtual_machine_template,
+             appliance_type: appliance_type,
+             tenants: [funded_t])
+
+      Atmosphere::Cloud::SatisfyAppliance.new(appliance).execute
+
+      expect(appliance.fund.reload).not_to eq funded_t.funds.first.reload
```
mkasztelnik: You can assign fund created in line 61 into variable, then you will not need to traverse over relations.

mkasztelnik:
Some small remarks in the code, but this change is quite important and I believe that is should be also tested manually before merging. Please let me know when it is deployed into vph-dev I will perform manual tests there.

app/providers/atmosphere/optimization_strategy/default.rb
nowakowski: Yeah, pretty much - it's used to determine how much a new appliance *would* cost (needed by UI) so it gets invoked without an appliance argument.

app/models/atmosphere/appliance.rb
nowakowski: OK

app/providers/atmosphere/optimization_strategy/default.rb
nowakowski: OK

nowakowski: Actually, `user_selected_tenants` is also used by `restrict_by_user_requirements`. Keeping it.

app/services/atmosphere/cloud/satisfy_appliance.rb
nowakowski: OK

app/services/atmosphere/cloud/satisfy_appliance.rb
nowakowski: OK

spec/services/atmosphere/cloud/satisfy_appliance_spec.rb
nowakowski: You're right; this spec no longer makes sense. Removed it.

spec/services/atmosphere/cloud/satisfy_appliance_spec.rb
nowakowski: OK

spec/services/atmosphere/cloud/satisfy_appliance_spec.rb
nowakowski: Yup, it's not necessary.

spec/services/atmosphere/cloud/satisfy_appliance_spec.rb
nowakowski: Right you are.

nowakowski:
@mkasztelnik Deployed on vph-dev and appears to work.

mkasztelnik:
@nowakowski are you sure that this branch id deployed into vph-dev? I see following entry in `Gemfile.lock`

```
remote: git://github.com/dice-cyfronet/atmosphere.git
  revision: 4399129fdb45d71920a5e9e448452bc5fcc67925
  branch: master
```

nowakowski:
OK; I failed to commit `Gemfile.lock` prior to pushing my branch to vph-dev. Fixed. Sorry.

mkasztelnik:
:+1:

mkasztelnik:
Unfortunately I need to remove :+1:, since not all remarks from my code review were taken into account. E.g. `Appliance#default_fund` method is not removed, in the spec private methods are used (e.g. `appliance.send(:default_fund)` in `appliance_spec.rb` - this spec can also be removed since `default_fund` will be removed.

nowakowski:
@mkasztelnik All done.

tbartynski:
:+1:

mkasztelnik:
:+1:

PR #243, opened 2015-08-26
Nil hourly cost immunity

tbartynski:


spec/lib/atmosphere/utils_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'spec_helper'
 require 'atmosphere/utils'
-#$: << '../../lib'
+# $: << '../../lib'
```
mkasztelnik: Why this line is needed? Maybe we can remove it?

mkasztelnik:
One small question and then :+1:

mkasztelnik:
What is more are you able to squash these 6 commits into one?

spec/lib/atmosphere/utils_spec.rb
tbartynski: I didn't add this line. I only corrected it's style.

mkasztelnik: I understand. If it is not needed then I propose to remove it.

tbartynski: It was not needed. Removed it.

PR #244, opened 2015-08-27
Verify that clew correctly hides appliance types which have VMTs only on inactive tenants

### LINKED ISSUE #190: Bug in flavor selection process when some compute sites are flagged as inactive
The AT flavor selection procedure invoked by CLEW APIs bugs out with a 500 error when there are ApplianceTypes which have a nonempty list of VirtualMachineTemplates but all of these templates reside on inactive compute sites:

`atmosphere/clew_appliance_types_serializer.rb in map_flavor at line 39
NoMethodError: undefined method id for nil:NilClass`

nowakowski:
Fixes #190


nowakowski:
This had already been fixed but I added a test which checks for the error condition.

mkasztelnik:
:+1:

Issue #75, opened 2015-01-08
Method new_vms_tmpls_and_flavors_and_tenants in manual optimization strategy should check if optimization policy params are present


mkasztelnik:
```
I, [2015-01-08T12:45:12.947823 #26651]  INFO -- :   Parameters: {"appliance"=>{"configuration_template_id"=>"214", "appliance_set_id"=>"536", "params"=>{}, "name"=>nil, "user_key_id"=>"39", "dev_mode_property_set"=>{"preference_cpu"=>"0", "preference_memory"=>"0", "preference_disk"=>"0"}, "compute_site_ids"=>nil}}
I, [2015-01-08T12:45:13.040284 #26651]  INFO -- : Completed 500 Internal Server Error in 92ms
F, [2015-01-08T12:45:13.897698 #26651] FATAL -- : 
NoMethodError (undefined method `[]' for nil:NilClass):
  vendor/bundle/ruby/2.1.0/bundler/gems/atmosphere-3dee9558043d/app/providers/atmosphere/optimization_strategy/manual.rb:21:in `new_vms_tmpls_and_flavors'
```

mkasztelnik:
It it even worse - manual optimization policy assume that there are `optimization_policy_params` available in appliance

@tbartynski can you explain why it was implemented in such a way and how we can use manual policy to start new appliance (not given in start new `appliance_set` request).

/cc @bwilk

mkasztelnik:
@tbartynski any deadline for this issue? It is quite important for ISMOP, what is more it blocks https://redmine.dev.cyfronet.pl/issues/3328

mkasztelnik:
@tbartynski any news here?

tbartynski:
This issue seems outdated. This code is already in master. See 

https://github.com/dice-cyfronet/atmosphere/blob/master/app/providers/atmosphere/optimization_strategy/manual.rb#L18

@mkasztelnik, @nowakowski please confirm it.

mkasztelnik:
It is not outdated. The scenario is as follow:

create new appliance set with e.g. type development and optimization policy set to true, next try to create new appliance with following params:

```
{ 
  "appliance":  {
    "configuration_template_id": "REAL_ID", 
    "appliance_set_id": "REAL_ID", 
    "params": {}, 
    "name": nil, 
    "user_key_id": "REAL_ID", 
    "dev_mode_property_set": {
      "preference_cpu": 0, 
      "preference_memory": 0, 
[...]

PR #246, opened 2015-08-27
checking for nil values in new_v__a_f_a_t method

tbartynski:
https://github.com/dice-cyfronet/atmosphere/issues/75

If used to fail if optimization_strategy_params or optimization_strategy_params['vms']
were nil.


mkasztelnik:
:+1:

Issue #245, opened 2015-08-27
While creating an appliance it should be checked if chosen opt strategy supports dev mode


tbartynski:
For instance manual strategy can not be used in dev mode. While instantiating an appliance object it should check whether chosen policy support dev mode.

tbartynski:
Create a class method in Default policy:
supports_dev_mode? that returns true.
Other implementations of optimization policies should extend Default policy and override if necessary that method.
Manual.supports_dev_mode? should return false.
While creating appliance it should be validated:
if appl.development && !appl.opt_policy.supports_dev_mode
  raise error
end

mkasztelnik:
I propose to generalize this method like this:

```
def support?(as)
  as.production?
end
```

tbartynski:
Ok. Any suggestions what error should be raised? Atmosphere::UnsupportedException?

mkasztelnik:
It was agreed that this check will be done as model validation

tbartynski:
Pull request ready for review.

PR #248, opened 2015-08-28
check if strategies support appl set of given type

tbartynski:
https://github.com/dice-cyfronet/atmosphere/issues/245


mkasztelnik:
:+1:

PR #257, opened 2015-09-08
enhanced fingerprint generation

### LINKED ISSUE #256: Enhance fingerprint generation for public keys

tbartynski:
https://github.com/dice-cyfronet/atmosphere/issues/256

If filename of the public key was manipulated running ssh-keygen command could be exploited
to inject some code to shell. Current invocation of ssh-keygen is safer.

Fix #256.


mkasztelnik:
:+1:

Issue #259, opened 2015-09-09
Ensure that only one ssh-rsa key is saved in UserKey model


tbartynski:
User was able to provide for instance two keys.

PR #260, opened 2015-09-09
sanitize public key content

tbartynski:
Fix https://github.com/dice-cyfronet/atmosphere/issues/259


mkasztelnik:
Please fix errors reported by hound

mkasztelnik:
I believe that regexp can be used here and first valid ssh public key can be taken. It should simplify code.

mkasztelnik:
:+1: but why `Gemfile.lock` is changed?

tbartynski:
Gemfile.lock modified by accident. Reverted this change.

PR #279, opened 2015-12-15
Optimize tenant, flavor and template selection

nowakowski:
Completely eliminates the n+1 query problem in optimizer.select_tmpl_and_flavor_and_tenant, thus fixing a nasty performance bug in PL-Grid Atmosphere deployment.

### DIFF
--- a/app/providers/atmosphere/optimization_strategy/default.rb
+++ b/app/providers/atmosphere/optimization_strategy/default.rb
@@ -90,8 +90,23 @@ module Atmosphere
         include Atmosphere::Utils
 
         def initialize(tmpls, appliance, options={})
-          @tmpls = tmpls
-          @appliance = appliance
+          @tmpls = VirtualMachineTemplate.where(id: tmpls.map(&:id)).
+                   includes(
+                     tenants: {
+                       virtual_machine_flavors:
+                       [:os_families, { flavor_os_families: :os_family }]
+                     }
+                   ).
+                   includes(:appliance_type)
+          if appliance.blank?
+            @appliance = nil
+          else
+            @appliance = Appliance.where(id: appliance.id).
+                         includes(appliance_set: [user: :funds]).
+                         includes(fund: :tenants).
+                         includes(:tenants).
+                         first
+          end
           @options = options
         end
 
@@ -139,7 +154,7 @@ module Atmosphere
           # in Atmosphere::Cloud::SatisfyAppliance
           # TODO: This is quite hacky; a better way to communicate errors to the user is needed.
           if best_template.blank?
-            best_template = tmpls.first
+            best_template = @tmpls.first
           end
 
           [
@@ -162,14 +177,13 @@ module Atmosphere
           if @appliance.blank?
             # This will happen in pre-instantiation queries (such as in the CLEW
             # "Start application" window) where no appliance is yet present.
-            tmpl.tenants.active
+            tmpl.tenants.select(&:active)
           else
             eligible_tenants = tmpl.tenants.active &
               @appliance.appliance_set.user.tenants
             if @appliance.fund.present?
               eligible_tenants = eligible_tenants & @appliance.fund.tenants
             end
-
             if @appliance.tenants.present?
               # If appliance is manually restricted to a specific subset of tenants,
               # return intersection of eligible tenants and appliance tenants.
@@ -187,7 +201,14 @@ module Atmosphere
           best_mem = Float::INFINITY
           opt_fl = nil
           mfs.each do |f|
-            present_cost = tmpl.get_hourly_cost_for(f)
+            flavor_os_families = f.flavor_os_families
+            os_family = flavor_os_families.detect do |family|
+              family.os_family_id == tmpl.appliance_type.os_family_id
+            end
+            present_cost = Float::INFINITY
+            if os_family.present?
+              present_cost = os_family.hourly_cost
+            end
             if present_cost < best_cost
               opt_fl = f
               best_cost = present_cost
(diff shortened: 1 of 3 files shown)


nowakowski:
@mkasztelnik Can you help out with this? It's quite urgent (btw., we probably need an "Urgent" label -- perhaps you could define one).

mkasztelnik:
Can you describe how can I help with it? What is missing? Only review or something more?

nowakowski:
@mkasztelnik I just need a code review.

mkasztelnik:
First of all can you squash all commits into one with meaningful name? Commits like "Interim commit." or "Woof." say nothing about what was changed and why.

app/providers/atmosphere/optimization_strategy/default.rb
```diff
@@ -105,7 +120,7 @@ def select
           best_tenant = nil
           best_flavor = nil
           instantiation_cost = Float::INFINITY
-          tmpls.each do |tmpl|
+          @tmpls.each do |tmpl|
```
mkasztelnik: 1 access method should be used in the whole code. So only `@tmpl` and remove `tmpl` getter or use `tmpl` getter. Using both is misleading.

app/providers/atmosphere/optimization_strategy/default.rb
```diff
@@ -187,7 +201,11 @@ def get_optimal_flavor_for_tenant(tmpl, t)
           best_mem = Float::INFINITY
           opt_fl = nil
           mfs.each do |f|
-            present_cost = tmpl.get_hourly_cost_for(f)
+            fofs = f.flavor_os_families
+            osf = fofs.detect do |fof|
```
mkasztelnik: Please use more meaningful names

spec/services/atmosphere/optimizer_spec.rb
```diff
@@ -115,8 +115,8 @@
                                   cpu: biggest_os_flavor.cpu,
                                   hdd: biggest_os_flavor.hdd,
                                   tenant: amazon)
-          biggest_os_flavor.set_hourly_cost_for(Atmosphere::OSFamily.first, 100)
-          optimal_flavor.set_hourly_cost_for(Atmosphere::OSFamily.first, 100)
+          biggest_os_flavor.set_hourly_cost_for(Atmosphere::OSFamily.first, 1)
+          optimal_flavor.set_hourly_cost_for(Atmosphere::OSFamily.first, 1)
```
mkasztelnik: Why this change is introduced?

app/providers/atmosphere/optimization_strategy/default.rb
nowakowski: Uh, I don't believe I understand. `@tmpls` is an instance variable which I'm iterating over. `tmpl` is a local variable which is only needed in the context of this specific loop - hence no `@`. What good would replacing `tmpl` with `@tmpl` do? Seems like a pointless waste of RAM.

app/providers/atmosphere/optimization_strategy/default.rb
nowakowski: OK.

spec/services/atmosphere/optimizer_spec.rb
```diff
@@ -145,6 +145,25 @@
...
+        vmts << create(
+          :virtual_machine_template,
+          appliance_type: atype,
+          tenants: [t]
+        )
+      end
+      t1 = Time.now.to_f
+      subject.select_tmpl_and_flavor_and_tenant(vmts)
+      t2 = Time.now.to_f
+      expect(t2 - t1).to be < 5
```
mkasztelnik: Please add error message to know what is wrong when this check fails (https://www.relishapp.com/rspec/rspec-expectations/docs/customized-message).

app/providers/atmosphere/optimization_strategy/default.rb
mkasztelnik: Take a look here https://github.com/dice-cyfronet/atmosphere/blob/optimize_tenant_selection_2/app/providers/atmosphere/optimization_strategy/default.rb#L171. That is why it is misleading.

spec/services/atmosphere/optimizer_spec.rb
nowakowski: The old selection logic was bogus. It preferred a flavor with price 100 on amazon over a flavor with price 80 on openstack even when both flavors satisfied AT requirements and the AT had VMTs on both tenants.

This was never caught because (1) the spec was also bogus (hence the change); (2) the situation never came up on PL-Grid.

Btw., I did not write the old selection logic. ;)

mkasztelnik: Can you point me to the place where amazon flavors are created? I'm trying to understand this spec, but it is quite complicated. I come into conclusion that to complicated factories are baaaaaaaaaaad, so many unclear dependencies and assumptions :cry:

app/providers/atmosphere/optimization_strategy/default.rb
nowakowski: You seem to be confusing `@tmpls` with `@tmpl`. Please re-read your original comment and make sure it doesn't contain any typos in variable names. :)

If you want me to remove `attr_reader :tmpls` then I'll do so. I don't think it's needed anymore.

mkasztelnik: You are right `@tmpl` should be `@tmpls` in my original comment - sorry for it - but the comment is still valid.

mkasztelnik: E.g. here https://github.com/dice-cyfronet/atmosphere/blob/optimize_tenant_selection_2/app/providers/atmosphere/optimization_strategy/default.rb#L245 and in other places `tmpls` getter is used.

nowakowski: Okay, I'll use the getter instead of referring directly to `@tmpls`.

spec/services/atmosphere/optimizer_spec.rb
nowakowski: Line 31 - `let(:amazon) { create(:amazon_with_flavors, funds: [fund]) }`.

IMHO default `lets` in specs are baaaaaaaaad but fixing this is out of scope of this PR. :)

spec/services/atmosphere/optimizer_spec.rb
nowakowski: Aye.

nowakowski:
I believe I'm done with all requested changes.

mkasztelnik:
:+1: from me. Please merge all commits into one before merge.

PR #280, opened 2016-02-02
Use sshkey lib to generate public key fingerprint

mkasztelnik:
Previously system command was used. This was OS specific and e.g. on Ubuntu 15.10 this does not work. Now `sshkey` library is used which should be agnostic.

### DIFF
--- a/spec/requests/atmosphere/api/user_keys_spec.rb
+++ b/spec/requests/atmosphere/api/user_keys_spec.rb
@@ -168,11 +168,8 @@ describe Atmosphere::Api::V1::UserKeysController do
         }
         post api("/user_keys", user), invalid_user_key_req
         expect(response.status).to eq 422
-        expect(json_response['details']).to eq(
-          {
-            'public_key' => ['can\'t be blank']
-          }
-        )
+        expect(json_response['details']['public_key']).
+          to include('can\'t be blank')
       end
 
       it 'returns 402 Unprocessable Entity status if invalid public key is provided' do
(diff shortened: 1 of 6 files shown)


spec/requests/atmosphere/api/user_keys_spec.rb
```diff
@@ -184,12 +181,8 @@
...
         post api("/user_keys", user), invalid_user_key_req
         expect(response.status).to eq 422
-        expect(json_response['details']).to eq({
-          'public_key' => [
-            'bad type of key (only ssh-rsa is allowed)',
-            'is invalid'
-          ]
-        })
+        expect(json_response['details']['public_key']).
+          to include('bad type of key (only ssh-rsa is allowed)')
```
tbartynski: Are you sure that hash equals can be replaced with include? The latter does not verify that other errors are not present? Just wondering if it's a good practise...

mkasztelnik: `include` checks only for selected element in array (https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/include-matcher). The problem I find here is that "is invalid" was sometime added, other time doesn't into error list and "is invalid" is not a case in this test only information about supported key type.

tbartynski: I know the difference and that's why I am not sure sure if it's ok to replace one with another. I guess that test is supposed to check that response contains the given error message and optionally ensure hat other errors are not. My concern is about the optional part. If you don't care about checking whether other errors occurred then I'm ok with it.

mkasztelnik: As far as I understand this spec (you are the original author :smile:) we care only about concrete error here.

tbartynski:
:+1:

PR #283, opened 2016-03-15
Remove VM only when it exists, do nothing otherwise

mkasztelnik:
Fixes problem with failing jobs, when trying to delete not existing VMs

### DIFF
--- a/app/workers/atmosphere/cloud/vm_destroy_worker.rb
+++ b/app/workers/atmosphere/cloud/vm_destroy_worker.rb
@@ -7,8 +7,8 @@ module Atmosphere
       sidekiq_options retry: 4
 
       def perform(vm_id)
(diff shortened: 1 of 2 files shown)


mkasztelnik:

tbartynski:
:+1:

Nuanda:
Is it possible to do that from the Admin UI?

nowakowski:
@Nuanda Individual VMs cannot be deleted from the admin UI (you can only remove an entire ApplianceSet, which will trigger `destroy` on all underlying VMs). This ticket, however, is about fixing an inefficient worker - not about improving the UI.

Nuanda:
I know - just asking.

nowakowski:
Oh, okay then. :)

PR #284, opened 2016-03-15
Possibility to configure url checker, timeout set to 1s

mkasztelnik:
Url checker timeout can be configured using:

```ruby
Atmosphere.setup do |config|
  config.url_monitoring.timeout = 123
end
```

@nowakowski, @tbartynski  please take a look


tbartynski:
:+1:

PR #287, opened 2016-06-17
flavors not existing in OS but associated with a vm are marked as ina…

### LINKED ISSUE #286: Flavors being used by a vm are not deleted (or made inactive) even if there are no longer available in OpenStack
When updating flavors Atmo does not delete notexisitng flavor if there a vm using it. It should at least mark it as inactive. Such a flavor will be used to instantiate new vm and will cause errors in OpenStack. Make sure this will not break billing.

tbartynski:
Flavors that exist in cloud but are used by a running vm are marked as inactive.

Close #286

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,6 +13,7 @@ gemspec
 group :development do
   gem 'quiet_assets'
   gem 'web-console', '~> 3.0'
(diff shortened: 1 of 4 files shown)


dharezlak:
:+1:

PR #291, opened 2016-09-05
JWT token handling is now possible

dharezlak:
JWT token should be passed via the `jwtToken` property. Also, `cloudFacadeUrl`, `username` and `roles` properties should be passed as usual. Do not pass the `csrf*` properties as they will take precedence. Accept this merge request if all works well.


nowakowski:
I need this merged (eventually) but there are failed checks and the PR is not flagged as ready for review. @dharezlak - could you comment on the security warnings? @mkasztelnik - what's our policy on merging branches which are not cleared by Hakiri?

dharezlak:
If relevant, I propose to create a separate issue for the security warnings which refer to server-side vulnerabilities. This PR consisted of only client-side (aka JS) code modifications.

mkasztelnik:
I agree with @dharezlak

PR #301, opened 2016-12-07
296 upgrade to rails 5

tbartynski:


mkasztelnik:
Migration should be also changed a bit:

```
DEPRECATION WARNING: Directly inheriting from ActiveRecord::Migration is deprecated. Please specify the Rails release the migration was written for:

  class RemoveHttpAndHttpsFromPmt < ActiveRecord::Migration[4.2] (called from load at /home/marek/.asdf/installs/ruby/2.3.1/bin/rake:23)
```

mkasztelnik:
Another element which can be easily fixed:

```
DEPRECATION WARNING: ActiveRecord::Base.raise_in_transactional_callbacks= is deprecated, has no effect and will be removed without replacement. (called from <top (required)> at /home/marek/git/atmosphere/spec/dummy/config/environment.rb:5)
```

It is defined in `config/application.rb`

mkasztelnik:
This is strange:

```
λ leeloo dummy → λ git 296_upgrade_to_rails_5* → rails s
=> Booting WEBrick
=> Rails 5.0.0.1 application starting in development on http://localhost:3000
=> Run `rails server -h` for more startup options
```

As far as I know webrick was changed into puma in rails 5. Why it is started here?

tbartynski:
See
https://devcenter.heroku.com/articles/getting-started-with-rails5#webserver
I've tried that and after adding puma to gemspec Webrick is not used any more.

PR #302, opened 2016-12-19
Remove zabbix and influxdb

mkasztelnik:
We are not using zabbix and influxdb, thus we can remove it.


db/migrate/20161214130933_remove_vm_monitoring_id.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveVmMonitoringId < ActiveRecord::Migration[5.0]
+  def change
+    remove_column :atmosphere_virtual_machines, :monitoring_id
```
nowakowski: Have you checked `db:rollback` on this migration?

mkasztelnik: Can you elaborate what is wrong here and why rollback can fail here?

nowakowski: According to http://edgeguides.rubyonrails.org/active_record_migrations.html#using-the-change-method you must supply a datatype when calling `remove_column` from within `change` (otherwise it's irreversible).

mkasztelnik: Good catch. Fixed in 19580b4

PR #304, opened 2017-02-28
Implement local PDP for Vapor integration

nowakowski:
Work in progress


nowakowski:
@mkasztelnik This functionality is meant to be utilized by https://github.com/VPH-Share/atmosphere-vph/pull/19 which represents "the other half" of this PR.

app/providers/atmosphere/local_pdp.rb
```diff
@@ -0,0 +1,54 @@
...
+      uat.present?
+    end
+
+    def can_start_in_development?(at)
+      Rails.logger.debug("can_start_in_development: using local PDP")
+      uat = UserApplianceType.where(
+          user: @current_user,
+          appliance_type: at
+      )
+      uat.present? and ['developer', 'manager'].include? uat.first.role
```
mkasztelnik: Don't use `and`, instead use `&&` (http://stackoverflow.com/questions/1426826/difference-between-and-and-in-ruby)

app/providers/atmosphere/local_pdp.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    def can_manage?(obj)
+      uat = UserApplianceType.where(
+          user: @current_user,
+          appliance_type: obj,
+          role: 'manager'
+      )
+      obj.user_id == current_user.id or uat.present?
```
mkasztelnik: Don't use `or`, instead use `||` (the same reason as for `and`).

config/initializers/pdp.rb
```diff
@@ -0,0 +1,2 @@
+# Initialize Atmosphere pdp class to DefaultPdp
+Atmosphere::at_pdp_class = Atmosphere::DefaultPdp
```
mkasztelnik: Are you sure we should put it here? If engine is included in dependent project there can be initializers order loading problem.

app/providers/atmosphere/local_pdp.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    def filter(ats, filter = nil)
+      ats.where(visibility_for_filter(filter.to_s))
+    end
+
+    private
+
+    attr_reader :current_user
+
+    def visibility_for_filter(filter)
```
mkasztelnik: It seems like copy paste from `DefaultPdp`. Instead `UserApplianceType` should be used to check which ATs are visible.

lib/atmosphere.rb
```diff
@@ -48,7 +48,7 @@ def self.admin_entities
   # manage.
   mattr_accessor :at_pdp_class
   def self.at_pdp(user)
-    (at_pdp_class || Atmosphere::DefaultPdp).new(user)
+    at_pdp_class.new(user)
```
mkasztelnik: Why default pdp was removed from this getter?

lib/generators/templates/atmosphere.rb
```diff
@@ -19,12 +19,6 @@
   #   main_app: [ AdditionalResourceViewsToShow ]
   # }
 
-  # PDP class for defining which Appliance Types user is able to start in
-  # development, production mode and which Appliance Types user is able to
-  # manage.
-  #
-  # config.at_pdp_class
```
mkasztelnik: If this option is not used anymore how other projects like plgrid or exp in ismop will be working?

spec/abilities/atmosphere/appliance_ability_builder_spec.rb
```diff
@@ -51,7 +51,7 @@
   def start_appl(type, at, pdp_method_sym = nil, pdp_response = nil)
     appl = build(:appliance, appliance_set: build_as(type), appliance_type: at)
     if pdp_method_sym
-      allow(pdp).to receive(pdp_method_sym).with(at).and_return(pdp_response)
+      allow_any_instance_of(Atmosphere::DefaultPdp).to receive(pdp_method_sym).with(at).and_return(pdp_response)
```
mkasztelnik: Please try to improve line 10 to return pdp double instead of allowing any action on DeafultPdp. Something like:

```ruby
let(:pdp) { double('pdp') } 
let(:pdp_class) do 
  pdp_class = double('pdp_class')
  allow(pdp_class).to receive(:new).and_return(pdp)
  pdp_class
end
let(:ability) { Atmosphere::Ability.new(user, true, pdp_class) }
```

spec/providers/atmosphere/local_pdp_spec.rb
```diff
@@ -0,0 +1,88 @@
+require 'rails_helper'
+
+describe Atmosphere::LocalPdp do
+  context '#can_manage?' do
+    let(:user) { build(:user, id: 1) }
```
mkasztelnik: Please don't use `id` in build and than saving it in db (in line 16) - this can lead to hard to find bugs.

spec/providers/atmosphere/local_pdp_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+    subject { Atmosphere::LocalPdp.new(user) }
+
+    it 'allows to manage owned appliance types' do
+      at = build(:appliance_type, user_id: user.id)
```
mkasztelnik: build is enough here? In `can_manage?` function you are doing db query, thus I'm not sure how it works since this AT does not exist in db.

spec/providers/atmosphere/local_pdp_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+
+    it 'allows to manage appliance types where management permissions have been granted' do
+      at = build(:appliance_type)
+      create(:user_appliance_type, user: user, appliance_type: at, role: 'manager')
+
+      expect(subject.can_manage?(at)).to be_truthy
+    end
+
+    it 'does not allow to manage unowned appliance types' do
+      at = build(:appliance_type)
```
mkasztelnik: Similar as above: AT does not exist in db so `where` from `can_manage?` function will always return empty set.

spec/providers/atmosphere/local_pdp_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+
+      expect(subject.can_manage?(at)).to be_falsy
+    end
+  end
+
+  context '#can_start_in_development?' do
+    let(:user) { build(:user, id: 1) }
```
mkasztelnik: Please don't use id in user build.

spec/providers/atmosphere/local_pdp_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+
+    it 'disallows starting production-access services in development' do
+      create(:user_appliance_type, user: user, appliance_type: at, role: 'reader')
+
+      expect(subject.can_start_in_development?(at)).to be_falsy
+    end
+  end
+
+  context '#can_start_in_production?' do
+    let(:user) { build(:user, id: 1) }
```
mkasztelnik: Please don't use id in user build.

mkasztelnik (changes requested):

mkasztelnik:
What about new created ATs, where is the place appropriate rights are added (UserApplianceType is created)?

lib/atmosphere.rb
nowakowski: To ensure that the `DefaultPDP` is not used when the request specifies otherwise (for example, when the PDP cannot be determined by analyzing the request). `DefaultPDP` is permissive, it would be a security risk to keep it as 'default' - safer to generate an exception.

app/providers/atmosphere/local_pdp.rb
nowakowski: I've prepared a brand new filter.

config/initializers/pdp.rb
nowakowski: This was meant to ensure backwards compatibility with legacy code which may expect to read this variable. I can remove it, but I'm not entirely sure what goes on e.g. in ISMOP etc. Perhaps safer to keep it here.

lib/generators/templates/atmosphere.rb
nowakowski: By relying on the new default initializer added in `pdp.rb`.

spec/providers/atmosphere/local_pdp_spec.rb
nowakowski: There's a special exception for services whose `user_id` attribute matches the requestor's ID (these are considered "owned" by the given user). This is also why the `User` object was built with a specific ID (but I've reworked it anyway).

spec/providers/atmosphere/local_pdp_spec.rb
nowakowski: Yes, but there's an exception for services flagged as owned by the current user - see line 36 in `local_pdp.rb`:

<pre>
      obj.user_id == current_user.id || uat.present?
</pre>

Since `at` is passed directly to the PDP method, this condition will succeed without running a DB query.

mkasztelnik:
Is it ready for second round?

nowakowski:
@mkasztelnik It is now! (Except for the stupid transient in fund selection - I will tackle that in a separate PR).

mkasztelnik:
Some small remarks/questions if tested locally I think we should push it into prod and improve when needed.

